PR #29, opened 2021-03-07
replace link by button to maintain scroll position

### LINKED ISSUE #19: maintaining scroll position

ziwot:
fix #19

fixed after the warning supplied by eslint: 

> jsx-a11y/anchor-is-valid: The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. If you cann... a bu
tton and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md

The added styles are here to cancel default button styles

Did not used reactstrap *<Button/>* element because it embeds a focus pseudo-class and it can't be overriden with styles specified via javascript

All the styles may need a great rework later, this MR is just a quick fix

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -16,6 +16,11 @@ button {
   border-top: 0;
   border-left: 0;
 }
+
+.nav button:focus {
+  outline: thin dotted;
+}
+
 /* a {
   color : #000
 }
--- a/src/components/AppNavbar.js
+++ b/src/components/AppNavbar.js
@@ -10,9 +10,10 @@ import {
   NavItem,
   NavLink,
 } from 'reactstrap'
-import { HicetnuncContext } from '../context/HicetnuncContext'
 import { Card, Col, Row } from 'reactstrap'
 import { keyframes } from 'styled-components'
+import { HicetnuncContext } from '../context/HicetnuncContext'
+import Button from './Button'
 import '../App.css'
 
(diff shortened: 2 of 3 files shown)


ziwot:
after some tests, there also a lost of scroll position when collecting, but this one is more complicated to solve 
(or maybe not, it's just the loading of nfts)

andrevenancio (commented):
this looks good. we should probably create a button component which inherits this styles. because there's a lot of these <a> through the repo that should really be just buttons. Do you mind updating this from <button> to <Button> and create that component that internally "resets" the button style, and receives a onClick?

ziwot:
@andrevenancio done, don't know if it's what you thought about.
I added a pseudo class in *App.css* to mimic current behaviour, when you click on sync, and come back from Kukai, the button is outlined (and pseudo classes really don't work when written in js).
I guess, a better way to have generic button style would be to use its `<Button>` element and extends Bootstrap styles, but it's a little more work

andrevenancio:
This will do for now. my idea is to create a button that can render either an <a> with a href, or a <button> if it only requires a onClick event. we can let that logic sit on the <Button /> component. but this will do for now. that way, we can keep the styling on one component instead of duplicated everywhere.

PR #34, opened 2021-03-07
Adding tags to Mint

andrevenancio:
* Adding tags to mint page.
* Removing unused properties.
* Re-ordering imports so its "globals" first, then "locals".
* removed unecessary binding since the methods in question are already arrow methods (which binds them to the class)


andrevenancio:
already got merged in another PR. closing.

PR #45, opened 2021-03-08
delimit tags by comma instead of newline

atomantic:
I noticed some tags make the page really long and it's easy enough to delimit with a comma


andrevenancio:
Today I was working on the page so this should be fixed as the layout now is slightly different.

PR #46, opened 2021-03-08
fixing react lint warnings

atomantic:


src/context/HicetnuncContext.js
```diff
@@ -22,8 +22,6 @@ export default class HicetnuncContextProvider extends Component {
 
       address: '',
 
-      op: undefined,
```
atomantic, line 25: this is duplicated. I find that alphabetizing object keys adds a nice way to make sure things are doubled and also helps code collaborators have fewer merge conflicts when adding keys :)

src/pages/about.js
```diff
@@ -22,23 +22,23 @@ export default class About extends Component {
...
+    //   listStyle: 'none',
+    //   right: '0',
+    //   top: '0',
+    //   marginTop: '15%',
+    //   marginRight: '25px',
+    //   fontFamiliy: 'roboto',
+    //   textAlign: 'right',
+    //   fontSize: '40px',
+    //   animation: 'fadeMe 1.2s',
+    // }
```
atomantic, line 41: these are unused...

src/pages/display.js
```diff
@@ -216,6 +217,7 @@ export default class Display extends Component {
                       </span>
                       <span onClick={this.collection}>
                         <a
+                          href="#collection"
```
atomantic, line 220: nice to be able to link to these anyway :)

src/pages/feed.js
```diff
@@ -42,17 +42,15 @@ export default class Feed extends Component {
...
             e.token_id !== 1131 &&
             e.token_id !== 1417 &&
             e.token_id !== 1418 &&
             e.token_id !== 1419 &&
             e.token_id !== 641 &&
             e.token_id !== 1547
-          ) {
-            return e
-          }
+          )
```
atomantic, line 53: same thing but always returns `true` or `false` instead of `true` or `undefined`

andrevenancio:
There's a few conflicts now. I've been doing a bit of cleanup myself but there's a lot of places that still need work. Its probably best to approach it with smaller PR's almost file by file.

andrevenancio:
Some conflicts here and some design changes being introduced since this PR was created. Here's a list from the top of my head so you could amend this PR accordingly.

- The idea is to go through each file and remove the unused properties. instead of commenting them out.
- there's a clear distinction between <a> <Link> and <button> an <a> tag opens an external page. a <Link> is used for route changes (using react-router-dom) and finally a <button> is used for onClick events. all of this is handled by src/components/button/index.js so you should use the <Button/> component.

If possible keep PR's as small as possible so its easier to avoid conflicts and quickly merge them. There's a lot of warning everywhere right now, it annoys me a bit, but we need to fix them page/component at a time.

atomantic:
I'll break this up into new pull requests :)

PR #47, opened 2021-03-09
Removed horizontal scroll

DanielFuqua:
I've removed the unnecessary overflow on the x-axis.

BEFORE:

Screen Shot 2021-03-08 at 8 39 33 PM

AFTER:

Screen Shot 2021-03-08 at 8 40 03 PM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "gitdoc.enabled": false
+}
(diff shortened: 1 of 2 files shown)


andrevenancio:
Thank you! we're doing a very page by page cleanup to avoid problems like this. but in the meantime we can merge this PR. Then, page by page, we should fix scrolling issues and eventually when everything is fixed, remove this patch. @crzypatchwork We can merge this one!

PR #48, opened 2021-03-09
Remove comment for security

neocybereth:
Not sure if it's the case but I assume you may not want this in the code. Feel free to close if not


src/pages/feed.js
```diff
@@ -82,62 +72,50 @@ export default class Feed extends Component {
...
-          style={{ maxHeight: '50vh', height: 'auto', width: 'auto' }}
-          src={`https://ipfs.io/ipfs/${
-            this.state.items[index].token_info.artifactUri.split('//')[1]
-          }`}
-          alt="💥"
-        />
-      </div>
-    )
+  setActiveTab = (tab) => {
+    console.log('yoo')
```
andrevenancio: 👋 Yo! but remove console log :)

neocybereth: :P will do!

neocybereth: Ah yeah looks like I accidentally pushed to master after submitting this, it's literally a one-liner. I'll fix that now

neocybereth: @andrevenancio now it's back in it's proper state :)

andrevenancio:
Not at all. Deeply appreciate the PR. I think it's probably a good idea to write a contribution.md where we document the thinking moving forward. The idea is to try to update the website to the point it's as optimised and easy to maintain as possible. Hence my many comments, I was trying to communicate some of the thinking to move ahead and it's only fair that is written down to help anyone to contribute! I'll look into that!

PR #51, opened 2021-03-09
Replace sync alert with Taquito syncing function

neocybereth:
I've replaced the alert that says "sync" and instead it now launches the wallet sync function (this.context.syncTaquito) to save the user from having to click the sync button after the closing the "sync" alert. Here's a video:

https://user-images.githubusercontent.com/14209806/110433936-051d4500-8116-11eb-93a4-ba400d355457.mov

### DIFF
--- a/src/pages/display.js
+++ b/src/pages/display.js
@@ -30,7 +30,6 @@ export default class Display extends Component {
     console.log(this.context.getAuth())
     await axios
       .post(process.env.REACT_APP_TZ, {
-        // 3.129.20.231
         tz: window.location.pathname.split('/')[2],
       })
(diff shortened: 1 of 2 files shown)


andrevenancio:
Simple and elegant! 👍 Will the behaviour replicate to other wallets?

neocybereth:
> Simple and elegant! 👍 Will the behaviour replicate to other wallets?

@andrevenancio yeah it will as this is just the taquito sync which is the base-layer for setting up the wallet. Here's a video:


https://user-images.githubusercontent.com/14209806/110523624-3e39d180-8177-11eb-871b-e962f0de6f37.mov

PR #52, opened 2021-03-09
Make artwork clickable

neocybereth:
This makes all videos and images clickable and takes you to their OBJKT page. So it's no longer just the OBJKT title that's clickable but also the artwork:

https://user-images.githubusercontent.com/14209806/110434948-4a8e4200-8117-11eb-8e6a-094b9d0589ce.mov

### DIFF
--- a/src/pages/feed.js
+++ b/src/pages/feed.js
@@ -100,36 +100,40 @@ export default class Feed extends Component {
     )
     const image = (index) => (
       <div>
-        <LazyLoadImage
-          className="media"
-          style={{ maxHeight: '50vh', height: 'auto', width: 'auto' }}
-          src={`https://ipfs.io/ipfs/${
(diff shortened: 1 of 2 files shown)


src/pages/feed.js
```diff
@@ -100,36 +100,40 @@ export default class Feed extends Component {
...
       <div>
-        <LazyLoadImage
-          className="media"
-          style={{ maxHeight: '50vh', height: 'auto', width: 'auto' }}
-          src={`https://ipfs.io/ipfs/${
-            this.state.items[index].token_info.artifactUri.split('//')[1]
-          }`}
-          alt="💥"
-        />
+        <a href={`/objkt/${this.state.items[index].token_id}`}>
```
andrevenancio: This should be a Link from react-router-dom as it refers to a route change.

neocybereth: Cool, I updated that @andrevenancio

PR #57, opened 2021-03-09
Separate the items that are for sale from the ones not for sale

neocybereth:
This PR separates the items into two categories: For Sale and Not For Sale

https://user-images.githubusercontent.com/14209806/110451618-3d7a4e80-8129-11eb-8c19-16a0db163ccd.mov


src/pages/feed.js
```diff
@@ -41,18 +44,9 @@ export default class Feed extends Component {
     await axios
       .post(process.env.REACT_APP_FEED, { counter: this.state.counter + 1 })
       .then((res) => {
+        const invalidIds = [1130, 1131, 1417, 1418, 1419, 641, 1547]
```
neocybereth: This just cleanups up that stuff below. Also do we know why there's an invalid **token_id**? B/c we can cut this issue off at the source if so.

djangobits: These were IDs that were removed manually as people started to upload porn and stuff. were currently talking about how to block stuff...

andrevenancio: I've suggested to @crzypatchwork that we could manually add a tag to those "invalid ids" instead of polluting our source code with an array that will likely just keep growing. If we agree on a tag like `adult` or something like that, I can exlude the images coming in the feed that have that tag. and then even if someone randomly lands on that page I can hide that "adult" tag from the list of tags. so its not displayed on the OBJKT detail.

src/components/ItemsForSale.js
```diff
@@ -0,0 +1,179 @@
+import React, { Component } from 'react'
```
andrevenancio: I'm slowly trying to create consistency across classes. The preference is now
1) Global imports
2) Local imports
3) CSS imports

src/components/ItemsForSale.js
```diff
@@ -0,0 +1,179 @@
...
+              color: '#000',
+            },
+          }}
+          href={`/objkt/${this.props.items[index].token_id}`}
+        >
+          OBJKT#{this.props.items[index].token_id}
+        </a>
+      </div>
+    )
+    const image = (index) => (
```
andrevenancio: I'm slowly trying to remove duplicated code from all the pages. I've only had time to do src/pages/objkt-display
But on that page you can see that there's a renderer that renders the correct media based on the mimetype. We should use that, and avoid having duplicates of image/video components everywhere. This is because we want to introduce more media types in the future, and we dont want to keep duplicating components around. Can you update so it uses the "renderMediaType" (https://github.com/hicetnunc2000/hicetnunc/blob/master/src/components/media-types/index.js)

src/pages/feed.js
djangobits: But the tags that currently exist are onchain, right (in the ipfs md)? They can't be updated IMHO, means we need a list in any way to add the IDs to block... in the discord channel someone made a proposal, that would be better than hardcoding it...

andrevenancio (commented):
Hi! I've added some components here and there. We were experimenting with "tabs" before and their design and we ended up going with a more minimalistic version that this "github" lookalike. We probably now need to build it as a component so it can be re-used here, but have a look in [here](https://github.com/hicetnunc2000/hicetnunc/blob/master/src/pages/objkt-display/index.js#L150)

src/pages/feed.js
andrevenancio: If you find it, do you mind sharing it here for visibility? I'm not sure if "we" (platform) can edit contracts, if we can't then we need a middle layer between the API request and the frontend response. and that should be server-side in my opinion.

neocybereth: I'm also looking into separating the displays into a grid to show like 4 artworks on a row (2 on mobile, 3 on iPad). Is this something you guys are interested in @djangobits @andrevenancio ? Figure I should see if it's even desired before building it all :D

crzypatchwork:
I think it would be interesting to consider the hDAO feed as well. one can address hDAOs to a NFT and hype it in the feed. this is a coming feature that could easily go with all this in the near future :)

src/pages/feed.js
andrevenancio: Yeah I think that's one of the things we want to do. I have build a media query mixin that we should use, but its located at `src/styles/layout.scss`. and this is an example on how you should use it `src/components/media-types/video/index.module.scss` in there we define a mobile first layout then desktop (but we should also introduce the tablet). Ping me on discord if you need me to point you in the right direction.

andrevenancio:
> I think it would be interesting to consider the hDAO feed as well. one can address hDAOs to a NFT and hype it in the feed. this is a coming feature that could easily go with all this in the near future :)

In terms of feature, yes. in terms of layour I prefer the simplicity we have on the user pages or on the OBJKT where the menu is a horizontal set of buttons where the enabled button is bold. But I'm happy if this feature gets merged and I'm happy to amend the layout later.

PR #62, opened 2021-03-09
fix react warning: emojis should be wrapped with aria label

atomantic:


andrevenancio:
I think this was a good idea! Wanna bring it back? @atomantic

atomantic:
I ended up just putting all of these into the one file update for feed. If you want them split as per item, I can do it that way too. I just realized late that my prettier plugin for VSCode was auto-formatting some other lines across two other pull requests :)

PR #66, opened 2021-03-09
fix react linting warnings and format using prettier

atomantic:
rolling these into one for this file. It looks like prettier formatting hasn't been run on the whole project so formatting on save adds a couple of other format updates.


andrevenancio:
Can you fix the conflict and edit this PR for `develop` not `master` :)

PR #72, opened 2021-03-10
Revert "Replace sync alert with Taquito syncing function"

crzypatchwork:
Reverts hicetnunc2000/hicetnunc#51


crzypatchwork:
cannot revert this branch. it changes the UX.

Issue #81, opened 2021-03-11
Multiple Prices Are Not Sorted


atomantic:
This object has 5 units, 2 are for sale: https://www.hicetnunc.xyz/objkt/3942

The first sale is 1 XTZ
The second is 5 XTZ

The code is simply grabbing the first item in the swap array and showing the 1/5 for sale for 5 XTZ rather than 1 XTZ.

The items should be ordered so the least expensive item is sold first.

```
[{swap_id: "3899", issuer: "tz1iyFi4WjSttoja7Vi1EJYMEKKSebQyMkF9", objkt_amount: "1", objkt_id: "3942", xtz_per_objkt: "5000000"},
{swap_id: "3876", issuer: "tz1iyFi4WjSttoja7Vi1EJYMEKKSebQyMkF9", objkt_amount: "1", objkt_id: "3942", xtz_per_objkt: "1000000"}]
```

something like:

```
const least_expensive = this.state.objkt.swaps.reduce((sum,item)=>{
        if(!sum || (Number(item.xtz_per_objkt) < Number(sum.xtz_per_objkt))) {
          sum = item;
        }
       return sum;
      })
```

crzypatchwork:
ty that is a initial feature for the secondary market

atomantic:
Thanks for merging the request. Looks like the site is currently down though due to an infrastructure change with the lambda cors settings: 

```Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://51rknuvw76.execute-api.us-east-1.amazonaws.com/dev/feed. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).```

PR #82, opened 2021-03-11
use lowest price for sale

atomantic:
Fixes https://github.com/hicetnunc2000/hicetnunc/issues/81

### DIFF
--- a/src/components/item-info/index.js
+++ b/src/components/item-info/index.js
@@ -4,6 +4,7 @@ import { Button, Primary, Purchase } from '../button'
 import { HicetnuncContext } from '../../context/HicetnuncContext'
 import { walletPreview } from '../../utils/string'
 import styles from './index.module.scss'
+import { lowestPrice } from '../../utils/lowestPrice'
 
 export const ItemInfo = ({
   token_id,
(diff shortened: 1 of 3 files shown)


crzypatchwork:
the thing is that in addition to this, the swap condition in the UI is 'are you the owner of the objkt'? gotta change that to 'do you hold this objkt?'. 


https://api.better-call.dev/v1/contract/mainnet/KT1RJ6PbjHpwc3M5rw5s2Nbmefwbuwbdxton/tokens/holders?token_id=3942

also for cancelling a swap one should compare the issuer of a swap and the holders of the objkts.
https://github.com/hicetnunc2000/hicetnunc-api/blob/d878ab9c59fc67781ccd1a678b0db8597d6fa0d6/index.js#L72
```
{
issuer: "tz1iyFi4WjSttoja7Vi1EJYMEKKSebQyMkF9"
objkt_amount: "1"
objkt_id: "3942"
swap_id: "3899"
xtz_per_objkt: "5000000"
}
```

atomantic:
Not sure I follow. Are you saying, in addition to this fix, there's a need to add a check for the current synced wallet address vs the issuer? Can we do that one after this merge?

atomantic:
cleaned up with prettier formatter and squashed

PR #88, opened 2021-03-11
Deleted some unnecessary files, added more SASS variable support.

Devetec:


package.json
```diff
@@ -29,7 +29,8 @@
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
     "react-scripts": "^3.4.3",
-    "react-scroll": "^1.8.1"
+    "react-scroll": "^1.8.1",
+    "typescript": "^4.2.3"
```
andrevenancio: Remove typescript. 

What is react-scroll? why do we need it?

src/components/header/index.js
```diff
@@ -28,13 +28,13 @@ export const Header = () => {
 
           <div className={styles.right}>
             <Button onClick={context.syncTaquito} secondary>
-              <Secondary>sync</Secondary>
+              <Secondary className={styles.textcolor}>sync</Secondary>
```
andrevenancio, line 31: I'm not a fan of this API. the styling for the <Secondary /> button should be on the sass file of the button. src/components/button/index.module.scss The idea is to just import components and not worrying about how that variable gets defined.

src/components/header/style.module.scss
```diff
@@ -17,12 +18,17 @@
...
     justify-content: space-between;
+	color: $color;
 
     @include respond-to('desktop') {
       width: 50%;
       margin: 0 auto;
     }
-
+	
+	.textcolor {
```
andrevenancio: You defining this .textcolor but then on the component there's nothing using this class. Its just easier to define the color on the container, and put color: inherit on the css

andrevenancio:
I've added several comments. can you please have a look through them?

src/components/header/style.module.scss
Devetec: Will change!

package.json
Devetec: I have no idea what react-scroll is, but I didn't add it. To be on the safe side, I don't want to delete it in this PR.

Devetec:
Is this better?

src/components/header/index.js
andrevenancio: So the idea with <Secondary> is that it takes care of the layout for you. Do not pass classes into it. The button itself reads the variables.

andrevenancio:
Still some comments!

andrevenancio (approved):

PR #89, opened 2021-03-11
use ipfs.io instead of dweb.link so videos render

atomantic:
I think this was meant to be accepted but got overwritten in some refactoring.

Videos are still broken presently in Safari (e.g. https://www.hicetnunc.xyz/objkt/3007)

Screen Shot 2021-03-11 at 2 35 22 PM

Running this branch locally plays the video: http://localhost:3000/objkt/3007


andrevenancio (approved):
thanks!

PR #90, opened 2021-03-11
Fixed the bug where tags are outside of the view if there are too many

TDMW:
Adding flex-wrap; wrap makes sure the outer container wraps and tags go to a new line if required.

### DIFF
--- a/src/pages/objkt-display/index.module.scss
+++ b/src/pages/objkt-display/index.module.scss
@@ -16,6 +16,7 @@
 
 .tags {
   display: flex;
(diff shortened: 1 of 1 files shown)


andrevenancio (commented):
all good, but please keep the spacing. its seemslike you're using 4 spaces or tabs? there's an .prettierrc config on the project.

TDMW:
Completely missed that, didn't have prettier running as well. Fixed it. Thanks for letting me know!

andrevenancio (approved):
perfect! thanks

PR #91, opened 2021-03-11
Added spacing to grid rows so items have vertical spacing too

TDMW:
Currently there's no vertical spacing in the display grid so items touch each other. This adds the same spacings thats used horizontally to the rows so it looks better.


andrevenancio (approved):
👍

PR #93, opened 2021-03-11
bring back lowest price swapping

atomantic:
Looks like some kind of major revert happened that obliterated some of the merges.


andrevenancio:
I'm not sure what happen, but my understanding was that @crzypatchwork merged this by accident and then revert it back. Was this a feature he requested?

crzypatchwork:
yes it's intended to be the first version of the second market

PR #97, opened 2021-03-12
show sold out! message - and editions when no swaps are present

haustraliaer:
This one was annoying me - I think it's important to show when OBJKTS are sold out vs not for sale.

Also the editions should be present in this case.

I've left the editions enabled for when it's "not for sale" as that was cleaner and I couldn't think of any reason why they shouldn't show there as well?

### DIFF
--- a/src/components/item-info/index.js
+++ b/src/components/item-info/index.js
@@ -9,15 +9,21 @@ export const ItemInfo = ({
   token_id,
   token_info,
   swaps,
+  transfered,
   total_amount,
   isDetailView,
 }) => {
   const context = useContext(HicetnuncContext)
 
   const notForSale = swaps.length === 0
+  const soldOut = notForSale && transfered > 0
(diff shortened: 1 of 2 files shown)


src/components/item-info/index.js
```diff
@@ -37,7 +43,7 @@ export const ItemInfo = ({
               <Primary>{walletPreview(token_info.creators[0])}</Primary>
             </Button>
           </div>
-          {!notForSale && <p>Edition: {edition}</p>}
```
andrevenancio, line 40: What does it show on the button? Would it be nice to see the editions (10/10) (so you know how many were minted) and on the button see SOLD OUT and have it disabled?

crzypatchwork:
to be sold out I guess the issuer shouldn't hold any NFTs, meaning that they were all distributed.

andrevenancio (commented):
thank you!

haustraliaer:
Yep that was my feeling too - if it’s sold out it just shows how many were sold - thanks for merging!

PR #108, opened 2021-03-12
Basic a11y improvements

TDMW:
I noticed it's impossible to tab-navigate through the website due to the button reset that removed the outline. I don't know if this is on purpose, but it makes the platform unusable for people with impairments. I also started adding some aria labels. 

Let me know if this is appreciated and I could continue improving the accessibility of the platform a bit more.


src/styles/mixins.scss
```diff
@@ -6,10 +6,6 @@
   background-color: transparent;
   cursor: pointer;
   font-size: 16px;
-
-  &:focus {
-    outline: 0;
```
andrevenancio, line 11: Instead of removing the :focus completely, add a font-weight: bold below the outline. I think it still helps with accessibility while not making the layout terrible.

src/components/header/index.js
```diff
@@ -21,18 +21,18 @@ export const Header = () => {
     <>
       <div className={styles.container}>
         <div className={styles.content}>
-          <Button onClick={() => handleRoute('/')}>
+          <Button onClick={() => handleRoute('/')} aria-label="Back to home">
```
andrevenancio: This is super important, but I'm not sure I like the API. mainly because the button isn't rendering the injected props. so this will not be attached. You need to add {...props} on the `return ( <Button {...props}>` something like that.

TDMW: Fixed this. The prop is quite long now (ariaLabel) but it could be something like a11y. However that doesn't really say much.

src/styles/mixins.scss
TDMW: Not entirely sure if I agree. If you add bold it only highlights text. I can' see the menu button being focused or any other element that doesn't have text. I now use the text color CSS var for focus, meaning you see a black outline or white depending on light/dark mode. Way less intrusive. However, still not perfectly visible. Could also look at doing tab detection so tabbing users will see the original blue focus, but non-tabbing users wont; https://medium.com/hackernoon/removing-that-ugly-focus-ring-and-keeping-it-too-6c8727fefcd2

oppenheimer-:
i know this issue is closed but i'd really like add a few points to the subject.

1. Dealing with :focus to get rid of "terrible layout" in a respectful manner ([background information](https://css-tricks.com/the-focus-visible-trick/))
```
:focus:not(:focus-visible){
    outline: none
}
```

2. Another thing which bothers me alot is that the logo is a non-functioning button.
Please make it <a> working link to "/" home.

3. The sync-button, when already synced becomes the wallet adress. In this state it does not provide any function but still communicates it by having a pointer cursor. Please make it a profile link, again, with an <a> tag.

4. "creations" and "collections" should be in their very own route "/{wallet}/creations" for example.

Issue #86, opened 2021-03-11
componentWillMount needs to be removed


andrevenancio:
`componentWillMount` is being deprecated by React. Avoid using it.

guidoschmidt:
Just opened a PR #115, as I felt like this was a super easy start to get more into the already existing code 🎁 . Might be better to switch to function components using `useEffect` in the long run.

PR #115, opened 2021-03-13
EDIT: replace componentWillMount with componentDidMount

guidoschmidt:
Used #86 to get a bit more into the code, as I considered the issue a low hanging fruit 🙃 🍎.

Afaik the common replacement for deprecated componentWillMount is to just use
componentDidMount. However it might be nice to switch to React Hooks in the
future and rewrite the components as function components:

https://reactjs.org/docs/hooks-intro.html

### DIFF
--- a/src/pages/display/index.js
+++ b/src/pages/display/index.js
@@ -26,7 +26,7 @@ export default class Display extends Component {
     creationsState: true,
   }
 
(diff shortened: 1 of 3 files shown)


andrevenancio:
conflicts since this is all using react hooks. I'm going to close it.

PR #116, opened 2021-03-13
EDIT/pages/mint + components/media-types: basic preview of uploads on mint page

guidoschmidt:
Simple and first implementation of preview of uploads on the mint page

### DIFF
--- a/src/components/media-types/index.js
+++ b/src/components/media-types/index.js
@@ -42,3 +42,35 @@ export const renderMediaType = (token_info, interactive) => {
       return <UnknownComponent mimeType={mimeType} />
   }
 }
+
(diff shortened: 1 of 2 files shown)


andrevenancio:
Not sure about this... I'm not a fan of the whole duplication of the switch statement you've done. Is there something better we can do?

guidoschmidt:
You are right, the switch duplication was rather hacky/ugly. Just commited an update which is imo better

andrevenancio:
I like this. I'm going to merge it because I dont want your work to go to waste. I think you did a really good job, but I'm already refactoring the mint page on another branch to allow for validation. But I do want to get this in. thanks!

PR #118, opened 2021-03-13
Enabled glb viewer autoplay

leonardschoelch:
Enabled autoplay in the glb viewer so animations will play


andrevenancio (commented):
can you just amend this to go to `develop` not `main` ?

andrevenancio (approved):

PR #122, opened 2021-03-14
added wallet

djangobits:
copies stuff easily findable on the internet, except for the fotos


djangobits:
new one includes this

PR #135, opened 2021-03-14
This individual shouldn't be banned for 2 gray-area mints

veqtor:
We instead ban Objekts


veqtor:
Ninja fixed by @andrevenancio

PR #136, opened 2021-03-14
Feature/faq-mint-guide init

violetforest:
Starting with a basic FAQ mint guide


andrevenancio:
The burn is implemented. however, burned token still appears on your wallet. I need @crzypatchwork to update API to include the owners so I can reference each objkt that you created agains the burn wallet and filter the burned items.

PR #144, opened 2021-03-14
Sanitize objekts before filtering

veqtor:
Again this will break feed if we don't do this because we filter on fields that might not be present


veqtor:
this was ninja fixed by @andrevenancio

PR #148, opened 2021-03-15
Update index.js for /about

mathMakesArt:
Added a nbsp after the "email" button, as the current website reads "...get in touch by **email**or on **discord**" (note the lack of space between "email" and "or")

### DIFF
--- a/src/pages/about/index.js
+++ b/src/pages/about/index.js
@@ -62,7 +62,7 @@ export class About extends Component {
               &nbsp;
               <Button href="mailto:hicetnunc2000@protonmail.com">
                 <Primary>
-                  <strong>email</strong>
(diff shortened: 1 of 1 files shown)


mathMakesArt:
**[Here](https://hicetnunc.xyz/about)** is a link to the relevant webpage!

andrevenancio:
can you make it outside the button? you want to separate the button element, not the element inside the button. otherwise you're expanding the clickable button area.

mathMakesArt:
> can you make it outside the button? you want to separate the button element, not the element inside the button. otherwise you're expanding the clickable button area.

Thanks for pointing that out, my mistake! I see it's now been fixed by @Devetec so I'll close this PR now.

PR #151, opened 2021-03-15
Stop potentially DDoSing API

veqtor:
We might be ddosing the API


veqtor:
We are DDoSing the API :(

andrevenancio:
You're right on the final catch to return an object. but removing the catch on the other requests will force that catch to be called.

Having the catch on the other requests will allow one to fail and still try to go to the next one. so in theory if github is down it still call the feed. I'm happy to merge this, but as it is it will fail if any of the requests fails

andrevenancio (approved):

PR #155, opened 2021-03-15
Updated the display view to show user metadata

LeithBA:
image

I added "wallet" and "alias" to the page state.
wallet is equal to the public address string.
alias is equal to the shortened version of that same string.

I then fetch the user data using the tzkt api and the wallet address.
If the user changed their alias on tzkt.io, I change the value of the variable alias to the tzkt alias. If not, I do nothing.
I then change the value of balance (in state) to the tzkt balance)

Please feel free to edit my changes, I am not very react fluent


Arty2:
Although the balance of each wallet is publicly available on the tracker, displaying it on hicetnunc is not self-evident.
It is no statement of earnings or spent through hicetnunc. Would it make collectors make more informed “purchases” if they can see what the creator's wallet currently stores?

LeithBA:
I absolutely agree with you. The only reason I displayed balance is that there was already the empty balance placeholder so I just tried to follow the original dev intentions.

andrevenancio:
> I absolutely agree with you. The only reason I displayed balance is that there was already the empty balance placeholder so I just tried to follow the original dev intentions.

that would have been me. If it's publicly available why not display it?

src/pages/display/index.js
```diff
@@ -29,23 +31,38 @@ export default class Display extends Component {
 
   componentWillMount = async () => {
     this.context.setPath(window.location.pathname)
-    const currentWallet = window.location.pathname.split('/')[2]
-    console.log('current wallet', currentWallet)
+
+    fetch(`https://api.tzkt.io/v1/accounts/${this.state.wallet}`)
```
andrevenancio: can you move this API call to `src/data/api.js` ? or maybe tzkt.js is more appropriate. I'm trying to make the pages and the components as "dumb" as possible so we can easily unit test them. the logic of data handling and any type of filtering that needs to happen (or not) I would rather keep outside the page itself. Is that okay?

andrevenancio:
Also please edit this PR to go to `develop` not `master` I need to update the Contributing doc with this info. sorry if it's not clear.

src/pages/display/index.js
LeithBA: Sounds good to me, I'll do that as soon as I get home.

LeithBA:
Please hold on this. I did a mistake when commiting

LeithBA:
This should be ready to merge. Let me know if something needs to be fixed.  🎉
image

Devetec:
Honestly, it shouldn't display the amount of tez. People really don't seem to like the feature.

LeithBA:
I totally agree. I'll gladly remove it if @andrevenancio agrees. I just did not want to change the intended UX.
Displaying the tez amount does not serve any purpose (since people can always pay to add tez in their accounts or withdraw all teir tez at anytime). It might also scare new users that willl compare themselves to "bigger" artists.

andrevenancio:
I appreciate checking in with me, but Its not my decision to make. I think @crzypatchwork initial intention was to have it there. But I'm happy with whatever the community thinks its best.

andrevenancio (approved):

LeithBA:
Oh sorry I got that wrong 😅.

@crzypatchwork confirmed it was ok to remove the balance on discord so I'll do that asap.

LeithBA:
image

Balance is removed and icons are now dark theme compatible

LeithBA:
image

Conflict with hdao resolved! :tada:

veqtor:
Looks super-nice, maybe for a later PR but would be nice with profile image too

veqtor:
LGTM

veqtor:
I have tested this, very fast, no bugs found, looks very nice, ready to merge in my opinion

Screenshot 2021-03-18 at 13 31 51

PR #158, opened 2021-03-15
fix swapping (calculating edition with no current owners)

atomantic:


atomantic:
Alternative fix that also works is changing to `Object.keys(owners||{})`

atomantic:
but maybe this is a bigger problem with data sanitation since owners is being passed in as `undefined`

andrevenancio:
thanks!

PR #175, opened 2021-03-17
Add another copyminter

veqtor:
Another one


veqtor:
Dunno if it's okay to leave a trailing comma

Issue #173, opened 2021-03-17
Multilingual Support


Quasimondo:
With hic et nuc being as sparse as it is language is fortunately not a huge issue, still there is a huge part of the world who's not speaking English fluently. Since currently the amount of strings that need to be translated is still very small I think now would also be a great opportunity to allow for supporting an interface in different languages.

What would it take to make that happen? And who can help with translations?

violetforest:
What part of the platform do you propose is multilingual?

andrevenancio:
I'm happy to work out this feature on the website itself. But there are some elements that I dont think we can translate. Those are the blockchain fields (title, description).

What I propose:
1) We have a component where the user selects language (like we have the circle to toggle the theme)
2) We store the language on localStorage (like we're doing with the theme) and we load a translation JSON containing all the key pairs and relative translation (structure proposal below)
3) we force a window.location.reload() after setting the localStorage page reloads, reads from the localStorage and loads the correct translation.json


```json
{
  "menu-mint": "OBJKTs",
  "menu-mint-label": "(mint NFTs)",
  "menu-manage": "manage assets",
  "menu-about": "about"
  ... etc
}
```

The translations can live on our github page, like we're currently hosting the (o.json, w.json, b.json for blocklist objkt, wallets and ban respectively).

Should I do the first batch in english, and people can expand from that into their native language? @Quasimondo I'm assuming German can be done by you?

violetforest:
I can do spanish. 

Was going to do translation of the FAQ but I'm assuming people would just do google translate? Or is that not a reliable tool?

andrevenancio:
> I can do spanish.
> 
> Was going to do translation of the FAQ but I'm assuming people would just do google translate? Or is that not a reliable tool?

I'm actually hard. because the FAQ is constantly evolving. So if something changes in the English version we then need to ask everyone to update their own faqs versions. I'm not sure what's the best approach for that use case.

Quasimondo:
Thanks everyone for jumping onto this! So yes, we definitely need a solution that can keep pace with the continuous changes that happen on the site. And of course making additions or updates to the translation list should be as easy as possible.

As for the question what should be translated I would just prioritize it by what functions get used most to least. Another question that might be more something for the future is if or how translation proposals should be "vetted" or checked for correctness.

somaticbits:
I'd be willing to do a french translation if that's something useful.

andrevenancio:
[...]

PR #182, opened 2021-03-18
Feature/footer refactor

andrevenancio:
As suggested on Discord here's what I think we should do:

* The feed should **NOT** automatically load new items as the user scroll to the end of the page. Instead it should load on demand as the user presses a "Load More" button.

**Why?**
By implementing this change, we're able to move the footer to the end of the page. Which will be a huge improvement on mobile where the footer occupies quite a bit of available area.

On mobile the theme button overrides the text in the footer, and the only solution for that would be increase the size of the footer by introducing a row for buttons and a row for text which will further reduce the amount of available space for mobile.

Furthermore there's a feature request to add multi language [#173](https://github.com/hicetnunc2000/hicetnunc/issues/173) and that's another button that needs to go on the footer and therefore further accelerate the need to properly style the footer and potencially increase its height.

This proposal makes for a elegant solution plus removes a dependency on [react-infinite-scroll-component](https://www.npmjs.com/package/react-infinite-scroll-component)

One this PR gets merged, I can work on the ability to virtualise the rendered nft's. What do I mean by that? If the user loads 30, 60, 90, 120 etc they are rendering all those elements on the page. virtualisation will enable to only render the necessary cells to fill up the screen height, and then we're simply swapping the content of those html nodes.

But that's an improvement to this first version.

### DIFF
--- a/src/components/footer/index.module.scss
+++ b/src/components/footer/index.module.scss
@@ -1,9 +1,7 @@
 @import '../../styles/variables.scss';
 
 .container {
-  position: fixed;
-  left: 0;
-  bottom: 0;
+  position: relative;
   padding: 2px 0;
   width: 100%;
   text-align: center;
@@ -13,10 +11,4 @@
   transition: background-color #{$theme-duration} ease-out,
     color #{$theme-duration} ease-out;
   z-index: 2000;
-
-  // & > * {
-  //   :last-child {
-  //     font-size: 50%;
-  //   }
-  // }
 }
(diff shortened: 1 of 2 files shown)


crzypatchwork:
I find it interesting but hDAOs or random feeds would need to go in the same logic of implementation

andrevenancio:
That shouldn't be a problem. The code is the same. the only difference is that instead of loading automatically you press a button at the end of the page to load more. hDAO code is still there just need to get the green light from you(if it works on the API) to enable it! pm on discord to sync a deployment

haustraliaer:
I'm definitely in favour of this change - was one of the things I had on my list to implement.

PR #201, opened 2021-03-20
Fixes for Vector/HTML iframe query param names and templates

pichiste:
A couple of small fixes:
- The Vector/HTML iframes were adding the query param `author`, while the templates were referencing `creator`. I changed it so they both use `creator` since this is the term used in the metadata.
- Simplified the HTML template and added a comment so users are aware that external sources will be blocked (i.e. trying to load three.js or p5.js cdn)

### DIFF
--- a/src/components/media-types/html/index.js
+++ b/src/components/media-types/html/index.js
@@ -11,11 +11,11 @@ export const HTMLComponent = ({ src, interactive, preview, token_info }) => {
     [styles.interactive]: interactive,
   })
 
-  let _author_ = false
+  let _creator_ = false
   let _viewer_ = false
 
   if (token_info && token_info.creators[0]) {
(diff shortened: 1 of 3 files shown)


pichiste:
@andrevenancio just wanted to sanity check that changing the query param `author` to `creator` in the iframes is ok. Or are there artworks already minted using `author`? The SVG template on `main` shows `creator` so I'm assuming thats what it should be?

andrevenancio:
I think that might have been a typo from my end. I agree it should be creator to keep the names in sync

andrevenancio (approved):

andrevenancio:
Are you on Discord @pichiste ?

pichiste:
> Are you on Discord @pichiste ?

Hey yeah I'm on there as @pichiste

PR #215, opened 2021-03-22
use lowest price for current swap and price display

atomantic:
not sure why this was commented out. Opening back using the lowest price for the current swap. The current site behavior is to just grab the first swap in the array, which isn't always the oldest swap posted.


veqtor:
LGTM

andrevenancio:
@crzypatchwork I'm not sure if this was something you removed on purpose or not. What do you want to do with this PR?

atomantic:
closing as superseded by newer base: https://github.com/hicetnunc2000/hicetnunc/pull/347

PR #219, opened 2021-03-22
allow camera, microphone, geolocation access requests inside iframe (SVG/html)

atomantic:
- adding allow `camera, microphone, geolocation`, which will allow SVG/HTML art to request access to these devices

Sample art for testing:
- https://www.hicetnunc.xyz/objkt/10788
- http://localhost:3000/objkt/10788


andrevenancio:
We need to chat about this.. we're introducing HTML, we should move this functionalities away from SVG.

atomantic:
Sounds good. An html file in the iframe would benefit from this change as well if it wants to be able to request camera, etc. nothing specific in this merge for SVG :)

andrevenancio:
We need to sync this and the #208 in the same PR. We're waiting on @veqtor to have a look on how to parse zip files to make that available. and this can be merged right after.

msub2:
Adding in xr-spatial-tracking to the allow list as well would let people mint WebXR NFTs, which I'd be very interested in!

veqtor:
I think allow same origin might be unsafe

src/components/media-types/vector/index.js
```diff
@@ -34,7 +34,8 @@ export const VectorComponent = ({ src, interactive, preview, token_info }) => {
       <iframe
         title="hic et nunc SVG renderer"
         src={iframeSrc}
-        sandbox="allow-scripts"
+        sandbox="allow-scripts allow-same-origin"
```
veqtor: We CAN NOT have allow-scripts and allow-same-origin since that's basically the same as disabling the sandbox

veqtor: It would allow to access the injected wallet, possibly hacking or click-jacking and stealing funds

atomantic: missed that nuance. I'll do some testing tomorrow to see if the HTML version works with camera now rather than SVG as is. Pushed to remove allow-same-origin.

andrevenancio:
This feature is cool. Thoughts @crzypatchwork ? Do we want to allow webcam / audio to be added? It can lead to quite interesting experiences. Any security/privacy concerns anyone might have?

tarwin:
I imagine that if you allow camera/microphone etc you will be allowing for the domain yeah? And the domain being "cloudflare-ipfs.com" means that ANYTHING on that domain can then get access to these services, once you've allowed it for one artwork. So then anything uploaded to IPFS (and accessed through cloudflare CDN) gets access. Which ... seems like a problem.

atomantic:
@tarwin that's a good catch. When prompted for usage of camera/location/etc, the user can choose to allow permanently for the domain or one time for this page load. Maybe an addition to the UI to warn users about allowing it always is necessary...

tarwin:
image

@atomantic ...

I don't think permissions for camera or microphone work on a per-request basis. Once you have granted access you always have access unless someone manually turns it off.

Geolocation works in a different way I think, where you can get it only when specifically requested / approved using the Permissions API (https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API). From the docs it looks like browsers were thinking of rolling all permissions into this API but currently (since like Chrome 44?) it has only been Geolocation and Notifications.

veqtor:
What about accelerometer? Is this also possible?

atomantic:
Looks like on Safari, I get prompted every time I load it but on Chrome it's permanent :(

pichiste:
Does it make sense to consider all features we want enabled at once?

Here's the full list of available features:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy#directives

Could pare it down to something like this for example:
```
accelerometer
autoplay
camera
fullscreen
gyroscope
magnetometer
microphone
xr-spatial-tracking
```

pichiste:
> image
> 
> @atomantic ...
> 
> I don't think permissions for camera or microphone work on a per-request basis. Once you have granted access you always have access unless someone manually turns it off.
> 
> Geolocation works in a different way I think, where you can get it only when specifically requested / approved using the Permissions API (https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API). From the docs it looks like browsers were thinking of rolling all permissions into this API but currently (since like Chrome 44?) it has only been Geolocation and Notifications.

Re: camera, I think if you view it from the iframe embed on the site, it will actually show "hicetnunc.xyz". I tried also disabling camera for "cloudflare-ipfs.com" and it still works. Running locally this is what I see:

Screen Shot 2021-03-31 at 12 32 44 PM

atomantic:
@pichiste I think `fullscreen` and `autoplay` might be things to leave out

andrevenancio:
Hey guys, shifting this conversation to https://github.com/hicetnunc2000/hicetnunc/issues/459

pichiste:
Updates on the topic: camera and microphone, among other features, have been added to HTML OBJKTs here: https://github.com/hicetnunc2000/hicetnunc/pull/462

I think the consensus is that this shouldn't apply to SVGs.

@atomantic, regarding geolocation, this has been left out for the moment because it requires some more discussion around security. Please feel free to open a new PR to get the ball rolling if it's something you'd like to see added!

PR #228, opened 2021-03-23
grid view and other small fixes

nbieva:
grid button + grid view styles


nbieva:
infinite scroll/load more may need adjustments. Not sure if it has to be automatic or "on demand".
Update: As discussed, no infinite scrolling, but a "load more" button. --> so it should be ok with what was pushed.

src/components/header/index.js
```diff
@@ -29,23 +29,23 @@ export const Header = () => {
 
           <div className={styles.right}>
             <Button onClick={context.syncTaquito} secondary>
-              <Secondary>sync {context.activeAccount}</Secondary>
```
nbieva, line 32: I think this is a mistake. This does not have to be removed.

andrevenancio:
Quite a bit of changes. Which makes it hard to review as you're trying to do too many things. I would prefer if you could break down this PR into smaller localised changes instead of trying to do everything in the same pull request.

Some notes:
1) I see 2 new fonts have been introduced. Why was this necessary? Not saying you can't, but general layout changes like this should be discussed with more people. Was it because you're using a certain system that doesn't have Courier New? Was it because you personally prefer Courier Prime? The way you're importing the fonts is also inconsistent.

- One font via @import on the src/styles/main.scss. You are also importing all the available font weights from the google font cdn, but not all styles besides `regular` (400) and `bold` (700) are used on the site. So this seems unecessary.
- Another font imported via HTML. Why the different approaches? If you need just 2 icons you should just add them as SVG not import the whole font. Please see how someone implemented the user detail page where you have the twitter, website, github icons as svg.

2) We shouldn't apply styles directly to the <body> element like you're doing here. You should do how we did with all the other styling across the site. The styling should be done to a container <div> and be applied to that container div using `classnames` dependency and not accessing the classList prop directly. The way you did it its not necessarly incorrect in isolation, but it is incorrect in the context of this site where everything else has been done differently.

nbieva:
Hi! Thanks for the return.. Not surprised ;-) First time I work in a react app. I did what I could to help. 
Changing the font was for better legibility while keeping it very close to Courrier New.
No problem.. Again, thanks for your return!

PR #231, opened 2021-03-23
Added new FAQs

violetforest:
with images

can you proof read? @thegladscientist


thegladscientist (commented):
Looks good! Can probably change first line to to say "Below you'll find guides to some of the most commonly asked questions on Hicetnunc." since it covers way more than just minting now :)

Issue #207, opened 2021-03-21
PDF Media Support


raynt:
There was some discussion with **_devetec_** on Discord about the integration of PDFs into the site. Is this a possibility?

andrevenancio:
Sure. I'll mark it as feature request.

clemsos:
I can draft a PR for this and send you for review @andrevenancio ?

skenaja:
PDF now supported, closing.

PR #241, opened 2021-03-24
PDF support

clemsos:
**WIP** not ready to merge

fixes https://github.com/hicetnunc2000/hicetnunc/issues/207 

- [x] support pdf mime type
- [x] allow pdf upload
- [x] display pdf in an I frame

### DIFF
--- a/src/components/header/index.js
+++ b/src/components/header/index.js
@@ -11,6 +11,9 @@ import styles from './style.module.scss'
 export const Header = () => {
   const history = useHistory()
   const context = useContext(HicetnuncContext)
+  //context.setAccount()
+
+  //console.log(ls.get('activeAccount'))
 
   const handleRoute = (path) => {
     context.setMenu(true)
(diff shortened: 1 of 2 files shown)


githubbbie:
What's left to do?

clemsos:
Need to pick a pdf lib. There is react-pdf but it used Mozilla PDF.js that seems a bit bloated. 

Maybe better to use sth lighter or just directly the browser support with sth like 

```
<object data="sample.pdf" type="application/pdf" width="100%" height="100%">
      <p>Alternative text - include a link <a href="saample.pdf">to the PDF!</a></p>
  </object>
```

Any preferred way ?

clemsos:
Trickiest part is to prevent PDF exploits. Also maybe why pdf.js can be needed after all

andrevenancio:
Why do we need a PDF library? If you load a PDF on an iframe will it not display?

clemsos:
yes we can load in an iframe too

clemsos:
hmm... I force-pushed after rebasing with less commits and that seems to have closed the PR :/ 

I have added a simple iframe viewer https://github.com/clemsos/hicetnunc/tree/feature/pdf

@andrevenancio could you reopen plz?

githubbbie:
Makes sense

githubbbie:
Was this merged?

PR #242, opened 2021-03-24
Several small html-related fixes

pichiste:


src/constants.js
```diff
@@ -136,3 +136,7 @@ export const ALLOWED_MIMETYPES = Object.keys(MIMETYPE)
   .filter((e) => e !== MIMETYPE.GLTF) // disabling GLTF from new updates
 
 export const ALLOWED_FILETYPES = Object.keys(MIMETYPE)
+  .filter(k => {
+    return  k !== 'ZIP1' && k !== 'ZIP2'
```
veqtor: Hmmmm

pichiste: `ALLOWED_FILETYPES` was only used as a text label, but now made `ALLOWED_FILETYPES_LABEL` to disambiguate. 
https://github.com/hicetnunc2000/hicetnunc/pull/242/commits/656640ad6b189f3a5f3c08ad912516ff88a48167

PR #246, opened 2021-03-24
Load languages locally so we can test translations

flpvsk:
Just started to work on translations and noticed that languages are loaded from github. This PR moves the `languages` folder to `public`, makes a symbolic link `public/languages` (back to) -> `languages` and makes the `App` load language from the `public` folder instead of loading them from github.

### DIFF
--- a/public/languages
+++ b/public/languages
@@ -1,0 +1,1 @@
+../languages
--- a/src/App.js
+++ b/src/App.js
@@ -22,8 +22,11 @@ const App = () => {
   //
   useEffect(() => {
     const language = getItem('language') || setItem('language', 'en')
+    const langRoot = process.env.NODE_ENV === 'development' ?
+      '/languages' :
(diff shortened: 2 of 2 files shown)


andrevenancio:
Though I agree with you, this is super helpful to test locally, in production we dont want to use relative paths. The reason its pointing directly to github is because we can at any moment change the copy directly from Github without requiring a deployment. At the moment deployments are manual so it makes more sense for the time to have the option to come in, change it on github and see the changes reflected. I'll you update your PR to check if we're on localhost to point to local, if we're not to point to github then I can accept it, but leave the files in the same folder as they're now now in the public!

flpvsk:
Cool, PR updated.

andrevenancio (approved):

andrevenancio:
Thanks! Appreciate it!

PR #253, opened 2021-03-25
Be a little more lenient with connection abilities

veqtor:
Added an initial whitelist which adds a very small list of simple sources that can be used to connect to places that allow for interactive NFTs.
Not sure this works to split line like this, anyone?

### DIFF
--- a/src/utils/html.js
+++ b/src/utils/html.js
@@ -122,18 +122,24 @@ export function injectCSPMetaTagIntoHTML(html) {
     'afterbegin',
     `
     <meta http-equiv="Content-Security-Policy" content="
+    frame-ancestors
+      *;
     upgrade-insecure-requests;
     default-src
       'none';
     script-src
       'self'
       'unsafe-inline';
-    manifest-src
-      'self';
-    media-src
+    script-src-elem
       'self'
-      https://ipfs.infura.io
-      https://cloudflare-ipfs.com/;
+      'unsafe-inline';
+    script-src-attr
+      'self'
+      'unsafe-inline';
+    style-src-elem
+      'self';
+    style-src-attr
+      'self';
     img-src
       'self'
       data:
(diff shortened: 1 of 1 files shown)


veqtor:
This PR adds the ability to:

**Query**
- Tezos (better call dev & cryptonomic)
- Ethereum (Infura and Etherscan)
- IPFS (infura and ipfs.io)
- LibP2P (support for webrtc and wss, bootstrap nodes)
- WikiData
- The Graph
- Etherscan
- CoinMarketCap
- openweathermap.org
- Hic Et Nunc

**Use images from**
- IPFS (infura and ipfs.io)

**Use media (video, audio, etc) from**
- IPFS (infura and ipfs.io)

**Use fonts from**
- IPFS (infura and ipfs.io)
- Google Fonts

Furthermore, it adds security by disallowing navigate-to and a bunch of XSS, clickjacking and other exploiting techniques

Style sheets and scripts still need to be embedded in the ZIP, anything else would be insecure.

veqtor:
Before:
before

After
after_2

andrevenancio:
@quasimondo Anything to add to the list?

Quasimondo:
> @Quasimondo Anything to add to the list?

Whilst I would like to connect to things like flickr commons or internet archive, this might be a bit dangerous since both sites have APIs that allow for uploading, too. So for now it looks good to me.

fraguada:
Can we add tzstats? https://tzstats.com/docs/api
Would be in a similar category as better-call.dev.

veqtor:
I will add

> Can we add tzstats? https://tzstats.com/docs/api
> Would be in a similar category as better-call.dev.

Shvembldr:
I'm waiting for this pull-request more than anything else. It is very important to be able to identify the collector and give him new features in the collected nft

hturan:
This is wonderful, great work @veqtor! It'll really open the door for dynamic artwork on the platform 🙏 I can already think of a number of ways I'd love to use this.

> Not sure this works to split line like this, anyone?

Splitting lines inside of `content` attributes within HTML is totally fine 👍

andrevenancio (approved):

andrevenancio:
Sorry took so long to review this! Good job!

PR #277, opened 2021-03-28
Create html-p5js-template.zip

n1ckfg:


andrevenancio:
Thanks for this! I'm going to have to edit some of the things inside mainly because there's a mix of tabs and spaces. and references as `./`. and we noticed that IPFS doesn't seem to play well with those. I'm also removing the LICENSE and putting the credits on the README.md if you're okay with it?

n1ckfg:
Sure!

PR #278, opened 2021-03-28
feat: UI feedback on curate button

tarwin:
Added small feedback for curate button while the user is waiting for their wallet to respond.

Note: My first time trying React (Vue person) so I hope I followed the conventions correctly.


tarwin:
I screwed this up. I assumed the <Curate> button was not used. But it turns out it is used by `Burn`, so this commit screws burn up. Will work on fixing this tomorrow and resubmitting.

PR #279, opened 2021-03-28
Added UI feedback to Curate button

tarwin:
- Renamed the `Curate` button to `ActionButton` as this seemed like what it was used for.
- Added a `CurateButton` component for use with the curate action so we could give UI feedback when it is clicked.

This is my first stab @ React so I hope it follows the patterns correctly.


tarwin:
Sorry, just been told I did it all wrong - using old style React instead of hooks. I'll try again ...

PR #282, opened 2021-03-28
Fixed double instagram icon on user profiles

LeithBA:
I woke up to this small bug. Instagram icon was displayed twice in the user profiles 
image

So I just removed the second one
image


skenaja:
Fixes #287

PR #295, opened 2021-03-29
Autoplay Checks for Visibility Before Running Videos

jonobr1:
👋 

I noticed while browsing the feed and the artist page that there can be performance congestion when there are a lot of videos on one page all autoplaying. I'm submitting a PR that only "autoplays" the video once it's visible on the screen. I know that the contributing guides say not to drill props, but I couldn't think of a more efficient way to dispatch a scroll event to every `VideoComponent`.

More information on dangers of `autoplay` from the Mux team [here](https://mux.com/blog/video-autoplay-considered-harmful/).

Awesome work and community!

### DIFF
--- a/src/components/media-types/index.js
+++ b/src/components/media-types/index.js
@@ -72,7 +72,7 @@ export const renderMediaType = ({
     case MIMETYPE.QUICKTIME:
     case MIMETYPE.WEBM:
       url = preview ? uri : `${IPFS}${path}`
-      return <VideoComponent src={url} interactive={interactive} />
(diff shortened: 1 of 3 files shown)


andrevenancio:
This is awesome. thanks dude! I'm halfway implementing Intersection Observer which is better as it doesn't require the scroll event to check if an element is on screen. but this will be useful regardless!

jonobr1:
Yeah, obviously don't feel obligated to merge this. But, happy to help!

andrevenancio:
closing. this ended up being covered by the intersection observer update. thanks!

jonobr1:
Awesome!

PR #308, opened 2021-03-29
Slide Show Mode for the Random page

TerryMay:
* Added slide-show mode to random feed page
 * Added Button to switch feed from scrolling to slide show
   ** I'm not married to button iconography or placement, looking for suggestions
 * Changed logic to hide 'load more' and 'minting' messages when slide show is running
 * I used an inline style because I am still learning how css works here

Future Work
 * css fade in and out between changes
 * allow users to set the change interval
 * pause interval when a video is playing it's first loop
 * clean full screen mode


andrevenancio:
closing this PR. It has conflicts and the functionality doesn't seem to be production ready. Also making one feed unecessarly different from the others is not ideal. But I do like the idea of having a mode where you can get random art in a slideshow way. I would  put this against the community first via https://hicetnunc.featureupvote.com/ and if it gets enough traction then we could think about implementing it.

TerryMay:
Thank you for reviewing my pr.  I should have mentioned that I wanted
feedback more than anything. I will follow your advice and propose the
feature.  I did not know that url existed. Thanks again for taking the time
to look into it.

PR #354, opened 2021-04-01
Added data: and blob: to images/media in CSP + fullscreen fixes

pichiste:


src/utils/html.js
```diff
@@ -65,7 +65,7 @@ export async function unzipBuffer(buffer) {
     if (entry.buffer.length === 0 && entry.path.endsWith('/')) {
       type = IPFS_DIRECTORY_MIMETYPE
     } else {
-      type = mime.lookup(entry.path)      
+      type = mime.lookup(entry.path)
```
veqtor, line 68: ?

veqtor:
LGTM

src/components/media-types/container.js
```diff
@@ -18,17 +18,43 @@ export const Container = ({ children = null, interactive }) => {
   })
 
   const toggleFullScreen = () => {
-    if (!document.fullscreenElement) {
-      document.documentElement.requestFullscreen()
+    // iPhone doesn't support Fullscreen API
+    // but we can toggle fullscreen class manually
+    const iPhone = /iPhone|iPod/.test(navigator.userAgent) && !window.MSStream
```
andrevenancio: maybe iPad.. ? :)

andrevenancio: Also, don't do it in the toggle Fullscreen function. that function shouldn't be called at all.

Do it at the beginning. and if its IOS just hide the button. leave the rest of the code

src/components/media-types/container.js
```diff
@@ -37,8 +63,10 @@ export const Container = ({ children = null, interactive }) => {
 
   useEffect(() => {
     document.addEventListener('fullscreenchange', fullscreenChange, false)
+    document.addEventListener('webkitfullscreenchange', fullscreenChange, false)
     return () => {
       document.removeEventListener('fullscreenchange', fullscreenChange, false)
+      document.removeEventListener('webkitfullscreenchange', fullscreenChange, false)
```
andrevenancio, line 68: if webkit is not supported, why do we need this part?

src/components/media-types/container.js
pichiste: ipad works actually, its just iphone that doesn't work with fullscreen API. tested here

pichiste: i can move the check to the top, but fullscreen does "work" in that it will fill the screen, even if you still see the window chrome. thought this was better than completely disabling, or?

pichiste: Screen Shot 2021-04-01 at 11 47 45 AM

pichiste: ok cleaned up as you suggested

src/components/media-types/container.js
pichiste: webkit on desktop and ipad does work, its just iphone. need to listen to this extra event to make it work properly

pichiste:
@andrevenancio implemented the requested changes, this one is ready to go, thanks!

src/components/media-types/container.js
```diff
@@ -5,6 +5,8 @@ import classnames from 'classnames'
 import { HicetnuncContext } from '../../context/HicetnuncContext'
 import styles from './styles.module.scss'
 
+const iPhone = /iPhone|iPod/.test(navigator.userAgent) && !window.MSStream
```
andrevenancio, line 8: dont want to be stubborn, but don't we need iPad here too? https://stackoverflow.com/questions/9038625/detect-if-device-is-ios

andrevenancio: @pichiste ☝🏻

pichiste: iPad supports fullscreen API strangely, I tested it here! It's just iPhone that is the problem. I know it doesn't make any sense... here's a screenshot from my iPad:

IMG_0015

PR #369, opened 2021-04-02
Fix for Safari/Firefox fullscreen on Desktop

pichiste:


Pamicel:
Tested on safari 14.0 and FF 86.0 👍

my polite suggestion for a change is to use polyfills for `document.fullcreenElement` and `element.requestFullScreen`  at the top of `src/components/media-types/container.js` rather that polyfilling inside the code

Pamicel (approved):
Tested on safari 14.0 and FF 86.0 👍

Pamicel:
Tested the updated code, still good

andrevenancio:
minor conflict! don't know where that came from!

pichiste:
@andrevenancio fixed!

andrevenancio (approved):

PR #380, opened 2021-04-02
Request more detail in bug report template 

skenaja:
NB: I cherrypicked ebd8327 (as it was commited direct to `main`) in order to create this PR against `develop`

This updates the template to request that users include an objkt and/or wallet id in bug tickets to more easily track down issues (especially in upstream APIs)


andrevenancio:
Do you mind fixing the conflict? thanks for the update!

PR #401, opened 2021-04-04
Adding alert for case of Brave browser shields issue

### LINKED ISSUE #337: Add alert message for Brave blocking service worker in HTML preview
By default, Brave blocks the HTML preview's service worker, throwing the following error in the web console:
```
Uncaught (in promise) DOMException: Failed to register a ServiceWorker for scope ('https://hicetnunc2000.github.io/hicetnunc/gh-pages/html-preview/') with script ('https://hicetnunc2000.github.io/hicetnunc/gh-pages/html-preview/sw.js'): The user denied permission to use Service Worker.
```

There's nothing we can do about this on our end, but we should let the user know that their browser settings are blocking this feature from working. Maybe something like:

"Your browser settings are blocking this feature from working. Please enable service workers for this site, or use a different browser"

For Brave, the fix is to simply disable the shield for this site:

Screen Shot 2021-03-31 at 7 15 02 AM

pichiste:
Closed #337

### DIFF
--- a/gh-pages/html-preview/index.html
+++ b/gh-pages/html-preview/index.html
@@ -62,8 +62,9 @@
             }
 
             function onError(err) {
-              const msg =
-                'There was an error generating the preview. If you are using Brave browser, please turn off shields and try again.'
+              const msg = navigator.brave
+                ? 'There was an error generating the preview. If you are using Brave browser, please turn off shields and try again, or use a different browser.'
+                : 'There was an error generating the preview. Please check the web console for errors.'
               alert(msg)
               reject(err)
             }
--- a/gh-pages/html-preview/index.html
+++ b/gh-pages/html-preview/index.html
@@ -62,6 +62,9 @@
             }
 
             function onError(err) {
+              const msg =
+                'There was an error generating the preview. If you are using Brave browser, please turn off shields and try again.'
+              alert(msg)
               reject(err)
             }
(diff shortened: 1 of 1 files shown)


gh-pages/html-preview/index.html
```diff
@@ -62,6 +62,9 @@
             }
 
             function onError(err) {
+              const msg =
```
andrevenancio: Can we add a check against the userAgent to check for `brave`? I think something like `navigator.userAgent.IndeOf('Brave') !== -1` might do the trick

andrevenancio: so it doesnt look weird for non Brave users. For non Brave users it might be more usedul to do alert(err) so if an issue actually happens, people can add the error into new issues. or a console.log(err) for both cases? whatever you think its best.

pichiste: Good call, just pushed an update that does this:
```javascript
const msg = navigator.brave
  ? 'There was an error generating the preview. If you are using Brave browser, please turn off shields and try again, or use a different browser.'
  : 'There was an error generating the preview. Please check the web console for errors.'
alert(msg)
```

The error already prints to web console so no need to add here.

andrevenancio:
You made me download Brave just to check if `navigator.brave` works 🤣

Issue #403, opened 2021-04-04
Title typo: hic en nunc


scalemodal:
![image](https://user-images.githubusercontent.com/81647978/113516110-d198d380-9578-11eb-8d82-6dcf9c5c7911.png)

Google

![image](https://user-images.githubusercontent.com/81647978/113516219-508e0c00-9579-11eb-8eec-98452bf0844b.png)

mescalina:
lol

andrevenancio:
Just by judging from partial images this is what I think it's happening.

Google is indexing our platform with the correct title. The feed's homepage is the "latest feed" vs the other types of feed "hDao" or "random", which is displaying what it should display.

Now the screenshot saying display was taken from "someone"'s wallet. hicetnunc.xyz/tz/somethingsomething which is also correct.

Now since you haven't shared more of that screenshot I'm not sure if that was the page you were in or not. (It doesn't matter). There's this thing called Server Side Rendering on Single Page Applications which they generate the HTML based of the first page the user tries to access. The html is injected and hence you see that "title Display" showing. After than, any page you move to just changes the DOM elements of the page using javascript but that initial (server side rendered) html will still show "display" regardless of which page you are in.

So, everything is as it should be.

Thanks for the report though.

andrevenancio:
Toke me a while to see that! You are right!! Fixed. Awaiting deployment

PR #405, opened 2021-04-04
fix: update title text to fix typo

Anemy:
Addresses https://github.com/hicetnunc2000/hicetnunc/issues/403


andrevenancio (commented):
Seems like the change you're submitting is the same as it was before.

image

andrevenancio:
Saw it!

Issue #274, opened 2021-03-27
Implement imageUri on mint process


andrevenancio:
Implementing imageUri like we did with the HTML version will enable us to achieve a couple of things:

Speed up how fast assets load on the main feed and on artist profile. It will also allow us to add a cover image to video formats, and further optimise the site speed by displaying the cover image only if the video is in view.

tarwin:
For images this could be auto-generated in HTML on upload in JS. For videos it is more complicated, but could require that an image is also uploaded along with video.

veqtor:
> For images this could be auto-generated in HTML on upload in JS. For videos it is more complicated, but could require that an image is also uploaded along with video.

Would be nice to automatically generate a DASH manifest for videos, or just simply a preview video at a fixed, lower bitrate.

tarwin:
I'm looking at using FFMPEG.wasm to generate a preview video of a lower bitrate, fixed length and max size. And do the same for images.

andrevenancio:
> For images this could be auto-generated in HTML on upload in JS. For videos it is more complicated, but could require that an image is also uploaded along with video.

I'm tempted to request it regardless of the media. I dont want to end up having multiple forms for multiple media types. Having the thumbnail field like we have the title, description, tags would (in my opinion) be better.

andrevenancio:
But that lower bitrate video features sound good! Does it mean that we need to convert it at mint time, and upload 2 assets?

tarwin:
> But that lower bitrate video features sound good! Does it mean that we need to convert it at mint time, and upload 2 assets?

We should be able to create a new video in the browser before they even mint it using FFMPEG.wasm

pichiste:
Re: how this all gets saved in the metadata if we're trying to follow the [spec](https://tzip.tezosagora.org/proposal/tzip-21/):

Currently the `displayUri` field is used for HTML cover images, so I guess that should be the same across other media types. 

The `thumbnailUri` field is what is displayed in wallets and is meant to be thumbnail size. Maybe we can auto-generate this one too? Currently it's just a placeholder image.

Where would the low-bitrate video thumbnail reference be stored? Seems like `externalUri` makes the most sense but not sure.

andrevenancio:
[...]

PR #408, opened 2021-04-05
Implemented cover/thumbnail generation and upload

pichiste:
This PR adds an "Upload Cover Image" field on the mint page. For image OBJKTs, this field is hidden as the cover/thumb can be generated from image itself.

Currently two images are generated:
- cover: max 1200 x 1200 px compressed image
- thumbnail: max 350 x 350 px compressed image

These images are then uploaded to IPFS and references stored in the OBJK metadata in `displayUri` and `thumbnailUri` respectively.

These can then be used for display in the feed or elsewhere.

Preview videos can be a feature implemented on top of this (as suggested by @tarwin and @veqtor  https://github.com/hicetnunc2000/hicetnunc/issues/274). I'd suggest those references be stored in the `attributes` field as defined in this [spec](https://tzip.tezosagora.org/proposal/tzip-21/), so that those can be used when available, but we always have regular image assets referenced in `displayUri` and `thumbnailUri`.

Open questions:
- Currently the cover/thumb images are uploaded on their own. Should they actually be uploaded in a directory together with the original media? 
- Do we need the cover/thumb of an animated GIF to remain animated?
- Should there be enforced cover/thumb dimensions (i.e. square) for audio, 3D models, HTML, and SVG?

**Note: For HTML OBJKTs, users will now add the cover image via the upload UI, rather than specifying a meta tag. We should update that info in the documentation and templates upon release.**


andrevenancio:
Thanks for this @pichiste ! I think we can request this in all media types (audio, video, pdf's, svg, glb, etc). We might not use them all on our platform, but if they exist then we're able to preview them correctly in our wallets.

I think that the GIF for example, we might not use it and load the `gif` directly on the feed, but having the thumbnail as a required field of any NFT will allow us to:
1) Have an image to share on social media every time someone advertises their NFT on Twitter or Facebook etc,
2) Allow the wallet to display a thumbnail of the art instead of having the default black circle we currently have
3) accelerate the feed by serving the thumbnail instead of the full resolution image, etc.

In terms of forcing ratio, it depends. the glb viewer (3d viewer) is always a rectangle, but the image viewer (png, gif, jpg) can have any ratio. I guess we leave it to the users to upload the thumbnail in whatever ratio they want. we can provide a check and see if the uploaded file is a GLB then we only allow squared images to be displayed, and we can implement a error if they try to upload any image for a GLB that is NOT squared. All other media types can have whatever ratio they want.

pichiste:
@andrevenancio That all sounds good, it's basically already implemented this way so nothing we have to change! (unless we want to enforce certain aspect ratios etc)

PR #414, opened 2021-04-05
Button specificity / Curate button feedback

tarwin:
refactor: made buttons more specific to function, not needing to be wrapped by <Button>

refactor: changed <Curate> button to <ActionButton>

feat: added a <CurateButton> specifically for the hDAO voting button which gives feedback (<Loader>) while waiting for wallet


tarwin:
Still screwed it up sorry.

PR #415, opened 2021-04-05
Button specificity + Curate button feedback

tarwin:
refactor: made buttons more specific to function, not needing to be wrapped by <Button>

refactor: changed <Curate> button to <ActionButton>

feat: added a <CurateButton> specifically for the hDAO voting button which gives feedback (<Loader>) while waiting for wallet


andrevenancio:
I dont mind this change. it requires you to have the same props duplicated on each button. I wonder if we should just destructure props into <Button..

Example:
```
export const Primary = (props) => {
....
return <Button {...props}>....</Button>
```

Or... probably better use `<Button primary /> <Button secondary />`

What do you think?

tarwin:
The `<Button primary />` makes sense to me. That said, should we also move things like the `<Curate />` button their own components? Was just trying to not screw with the codebase too much to start with.

peterschmidler:
I think this would fix: https://github.com/hicetnunc2000/hicetnunc/pull/415

PR #416, opened 2021-04-06
Make <INPUT>s clearer

tarwin:
feat: changed "AMOUNT" to "QUANTITY" for inputs as this makes more sense in English
feat: added a border around inputs to make them clearer, especially when using labels instead of placeholders
feat: changed <Inputs> to use labels as it is bad practice to only have placeholders and can be confusing and lead to mistakes

It was too easy to make mistakes when minting or swapping. Part of this problem is the fact that only "placeholders" are used in the inputs. I realise this was partly to keep the interface as minimal as possible but made the process of minting / swapping confusing.

I also changed the label "AMOUNT" to "QUANTITY" as this makes it clearer it is referring to the number of OBJKTs. AMOUNT, especially when there is a lack of titles / help text can be confused more easily with price I think. I could be totally wrong with this though.

### DIFF
--- a/src/components/input/index.js
+++ b/src/components/input/index.js
@@ -3,25 +3,31 @@ import styles from './index.module.scss'
 
 export const Input = ({
   type = 'text',
-  placeholder = 'placeholder',
+  placeholder = '',
   name = 'input-name-not-set',
   min,
   max,
   maxlength = 500,
   onChange = () => null,
   disabled,
+  label,
+  helpText
 }) => {
   return (
-    <input
-      type={type}
-      placeholder={placeholder}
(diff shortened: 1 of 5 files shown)


andrevenancio:
I only saw this comment now. I do agree with you on the label! thanks

andrevenancio:
I've reverted this. a label is a good practice, but visually the form is not so great. I thought that you would be changing the placeholder text to be in a label, but visually it will be the same. I'm keeping the value feature though. Someone is reviewing the design, I dont wanna change it now until we have the final version.

tarwin:
Makes sense.

PR #417, opened 2021-04-06
Mint: Remember values

tarwin:
When coming back from Preview, remember values in inputs.

### DIFF
--- a/src/pages/mint/index.js
+++ b/src/pages/mint/index.js
@@ -120,12 +120,14 @@ export const Mint = () => {
                 type="text"
                 onChange={(e) => setTitle(e.target.value)}
                 label="TITLE"
+                value={title}
               />
 
               <Input
                 type="text"
                 onChange={(e) => setDescription(e.target.value)}
                 label="DESCRIPTION"
+                value={description}
               />
 
               <Input
@@ -133,6 +135,7 @@ export const Mint = () => {
                 onChange={(e) => setTags(e.target.value)}
                 label="TAGS"
                 helpText="tags (comma separated. example: illustration, digital, crypto)"
+                value={tags}
               />
 
               <Input
@@ -141,6 +144,7 @@ export const Mint = () => {
                 onChange={(e) => setAmount(e.target.value)}
                 label="QUANTITY"
                 helpText="number of OBJKTs to mint"
+                value={amount}
               />
 
               <Input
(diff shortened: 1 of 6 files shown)


src/pages/mint/index.js
```diff
@@ -119,34 +119,42 @@ export const Mint = () => {
               <Input
                 type="text"
                 onChange={(e) => setTitle(e.target.value)}
-                placeholder="title"
```
andrevenancio, line 122: Is there a reason why you're changing the <Input /> API and removing the placeholder text? with API changes like this I would like to get a rational on why we need to change it?

src/pages/mint/index.js
```diff
@@ -119,34 +119,42 @@ export const Mint = () => {
...
+                label="TITLE"
+                value={title}
               />
 
               <Input
                 type="text"
                 onChange={(e) => setDescription(e.target.value)}
-                placeholder="description"
+                label="DESCRIPTION"
```
andrevenancio, line 129: also, including new text will force me to chase everyone who contributed towards the translations. why do we need this?

src/pages/mint/index.js
```diff
@@ -119,34 +119,42 @@ export const Mint = () => {
...
                 onChange={(e) => setAmount(e.target.value)}
-                placeholder="amount"
+                label="QUANTITY"
+                helpText="number of OBJKTs to mint"
+                value={amount}
               />
 
               <Input
-                type="royalties"
+                type="number"
```
andrevenancio, line 151: This I like! Was that a bug? oops.

tarwin: Yeah, I'm guessing it was a bug :D

src/pages/mint/index.js
tarwin: Good question. "Placeholder" has a specific meaning in HTML inputs, which is what it was used for before. You can still use placeholder how it was used earlier. I just added `label` which adds a `<Label/>` around the input.

src/pages/mint/index.js
tarwin: I'm happy to chase down the translations - I should just look at who commited translation changes right? I wasn't able to work out how the translations work actually so thought maybe they won't being used yet.

andrevenancio:
This is a duplicate of the PR that got submitted and merged https://github.com/hicetnunc2000/hicetnunc/pull/416

PR #434, opened 2021-04-07
Several fixes/cleanup in HTML preview and sandbox page + CSP fix

### LINKED ISSUE #411: Content Security Policy issue when using DRACOLoader
Hello :wave:
I'm trying to publish an interactive game made with Three.js and I'm getting errors in the preview. The game loads a heavy GLTF model and uses the DRACOLoader for that (see https://threejs.org/docs/#examples/en/loaders/DRACOLoader). I suspect that the loader uses workers under the hood. And it seems that hicetnunc is blocking access to those workers with the error message

> Content Security Policy: The page’s settings blocked the loading of a resource at blob:https://hicetnunc2000.github.io/39efd702-2e43-e049-8227-ca769b91e0ae (“default-src”).

I think that the path to the worker has a slash in the beginning and that is why the complete URL is not correct and hence is being blocked. Any ideas what can be done in this case? I wonder if I am the first to use the official DRACOLoader in a project on hicetnunc :see_no_evil:

**To Reproduce**
Use the DRACOLoader to load a draco compressed model. Continue as per official documentation.
```
var dracoLoader = new THREE.DRACOLoader();
dracoLoader.setDecoderPath(''); // everything is in the root folder
dracoLoader.setDecoderConfig({type: 'js'}); // optional
```

**Expected behavior**
The compressed model should load and be encoded by the draco loader.

**Screenshots**
Screenshot 2021-04-05 at 19 41 07

 - macOS Big Sur 11.2.2 (20D80)
 - Firefix 88.0b7 (64-bit)

Please help me figure something out 🙏

**UPDATE**: After digging further I think I have found the line that creates the blob url inside the DRACOLoader class: https://github.com/mrdoob/three.js/blob/a382159ab8575857bc042d314019bf45edb14117/examples/jsm/loaders/DRACOLoader.js#L313

pichiste:
- Fixes double error alert bug by moving index.html validation out of sandbox page
- Improves error messaging in HTML preview
- Closes #411, issue about CSP blocking blob worker-src


andrevenancio (approved):

andrevenancio:
just a small conflict! all seems good!

pichiste:
@andrevenancio resolved!

Issue #469, opened 2021-04-10
p5.sound is blocked by the sandboxed iframe because it tries to load a script from a blob src


m0wh:
Hello :)
Not reeeealy a bug (it's expected in a way), but I'm trying to mint a p5 sketch with some p5.sound stuff, but the preview just displays "Loading...". The sketch works as expected locally and I'm not using any cdn, everything is in my zip file, but p5.sound tries to load something from a blob src witch is disallowed by hicetnunc's sandboxed iframe.

1. Compress a folder with an index.html importing local p5.min.js, p5.sound.min.js and a script using these libraries.
2. Preview it on hicetnunc with your console open

![image](https://user-images.githubusercontent.com/36423190/114268399-261cd280-9a01-11eb-9216-3f8490a1f760.png)

Thanks :)

P.S. If there is no easy solution to this, I will rewrite my p5.sound part with vanilla js :)

originaltomato:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src  is not supported in iOS and child-src seems like only full solution

clayheaton:
This issue is fixed per this bug:

https://github.com/hicetnunc2000/hicetnunc/issues/1177

@crzypatchwork I believe this can be closed, too.

PR #475, opened 2021-04-11
Added blob to script-src in CSP

pichiste:
Some often-used packages/libraries use blobs as script sources. 

For example, the p5.sound library: https://github.com/hicetnunc2000/hicetnunc/issues/469

This simply enables `blob` in 'script-src` of CSP.


veqtor:
LGTM

andrevenancio (approved):

PR #492, opened 2021-04-13
Feature/masonry

violetforest:
CSS for horizontal masonry layout. 4 column on desktop, 3 on tabet, 1 on mobile.

This looks a lot nicer because theres no gaps due to differign heights but the fallback is that it orders things from top to bottom instead of left and right. But its worth considering.

Screen Shot 2021-04-14 at 12 28 46 AM

### DIFF
--- a/src/pages/display/index.module.scss
+++ b/src/pages/display/index.module.scss
@@ -26,7 +26,6 @@
 }
 
 .list {
-  column-count: 4;
   column-gap: 1em;
 
   @include respond-to('tablet') {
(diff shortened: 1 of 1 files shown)


andrevenancio (approved):
welcome to the frontend gang! 🎉

andrevenancio:
I'm just realising, you're not specifing the mobile/tablet right?

violetforest:
@andrevenancio  hmmm what do you mean? I'm usign the same media queries as the last version

```
@include respond-to('tablet') {
    column-count: 3;
 }

@include respond-to('desktop') {
    column-count: 4;
}
```

andrevenancio:
I was looking at the wrong place. sorry

PR #505, opened 2021-04-15
Use OBJKT info to set tab visibility based on user role

### LINKED ISSUE #494: Swap and Burn options Unavailable for Secondary Owners?
**Describe the bug**
Swap creation and burning appear to be unavailable for secondary owners (owners of OBJKTS which were created by someone else).

**To Reproduce**
1. Purchase an NFT on HicEtNunc (or find one in your collection that you already own)
2. View that NFT's OBJKT page
3. Regardless of wallet connection status, the "Swap" and "Burn" options are unavailable. Only "info" and "collectors" tabs appear.

**Expected behavior**
Additional "Swap" and "Burn" tabs should be present, with their own respective functionalities.

*To clarify, all of these functionalities are present for pieces that I created myself. It's only secondary pieces which have this problem.*

**Screenshots**

 - OS: Windows 10
 - Browser: Firefox
 - Version: 87.0 (64-bit)

**Additional context**
I have heard the same complaint from other users, so this issue is not just restricted to my own wallet.

Wallet address: tz1gsrd3CfZv4BfPnYKq5pKpHGFVdtGCgd71
OBJKT that I own and created myself, and so I can swap: https://www.hicetnunc.xyz/objkt/9953
OBJKT that I own but secondarily, and so I can not swap: https://www.hicetnunc.xyz/objkt/13147

skenaja:
Fixes #494

This will probably need conflict resolution to cope with the temporary workaround applied to `main` in 0597531f66fcb57783dbdfb8c1fab720911a6a84 which has not yet been merged back to `develop`.......

### DIFF
--- a/src/pages/objkt-display/index.js
+++ b/src/pages/objkt-display/index.js
@@ -14,8 +14,8 @@ import { Info, Collectors, Swap, Burn } from './tabs'
 const TABS = [
   { title: 'info', component: Info },
   { title: 'collectors', component: Collectors },
-  { title: 'swap', component: Swap, creatorOnly: true, secondaryMarket: true }, // visible if user is the creator or if user can sell on secondary market
-  { title: 'burn', component: Burn, creatorOnly: true, secondaryMarket: true }, // visible if user is the creator
+  { title: 'swap', component: Swap, visibility: ['creator','owner'] }, // visible if user is the creator or if user can sell on secondary market
+  { title: 'burn', component: Burn, visibility: ['creator','owner'] }, // visible if user is the creator or if user can sell on secondary market
 ]
 
 export const ObjktDisplay = () => {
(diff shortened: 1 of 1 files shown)


skenaja:
Screenshots (based on my wallet# and own objects):

As creator:


As owner:


With neither role:

src/pages/objkt-display/index.js
```diff
@@ -28,6 +28,21 @@ export const ObjktDisplay = () => {
 
   const address = context.acc?.address
 
+  var roles = []
+  function checkRoles(obj,addr) {
+    if (
+      Object.keys(obj.owners).length > 0 &&
```
andrevenancio, line 34: its easier to understand. the only issue is that the owners sometimes comes empty. And @crzypatchwork prefers to display it for everyone than risk not showing the swap (enabling the 2nd market) if that api fails.

skenaja: Sounds like HIC should calculate the owners itself from chain, will lessen reliance on timely data from upstream provider.

This is not just a refactor, it does correct the bug that swap tab was only showing to creator, when API was showing correct data.

I'll close for now, but I recommend it would be needed in future for usability, as it would help remove confusion from non-owners - the appearance to them of swap and burn buttons has been reported as a bug in Discord. Happy to resubmit when the API reliability is increased.

skenaja:
Closing as discussed in code review comments.

crzypatchwork:
this is an important bug fix.

the only problem is when the API is unstable. so we should treat such exceptions also.

PR #530, opened 2021-04-18
Add tags to display page

esnho:
I love this project and decided to add a way to filter user's creations.

The method can be easily ported on collection page too.

I have the feeling that to move the system in its own files can be useful to reduce the index file size.

animation


andrevenancio:
@crzypatchwork What do you think? I like this feature. Code wise all looks great! 🎉

oldominion:
Awesome. It would be great to be able to direct link to a specific tag/list of tags too. e.g. ../creations/tags?="video,pickles"

esnho:
> Awesome. It would be great to be able to direct link to a specific tag/list of tags too. e.g. ../creations/tags?="video,pickles"

Something similar is doable, I didn’t develop the feature to keep the PR small.

esnho:
A nightmare: @andrevenancio is possible that this filters out objkts without tags?
Probably I have to create a set of mock data to play with

esnho:
My nightmare was real, last commit manage creations without tags 🥶

Screenshot 2021-04-22 at 23 15 32

esnho:
@andrevenancio can you or someone else help me understand if this code works?

It assumes that a creation without tags have`token_info.tags.length === 0`

Is this correct?

andrevenancio:
I've seen cases where they are just an empty string and in that case I think it's better to tag them as if they had no tag at all. You could just copy a API response from the console and modify the data locally and test with that modified data for cases where there is no tags. And see if it works. Or you can write an unit test on the component and test data with a mix of different types of tags. Hope it helps

esnho:
I did a test with real data, apparently objkts without tags contains an array with a single empty string tag, now the code looks for that.

oldominion:
> > Awesome. It would be great to be able to direct link to a specific tag/list of tags too. e.g. ../creations/tags?="video,pickles"
> 
> Something similar is doable, I didn’t develop the feature to keep the PR small.

There's a PR with some of this functionality already submitted here by the way: https://github.com/hicetnunc2000/hicetnunc/pull/552 - it would need to work with properly with that format.

esnho:
> > > Awesome. It would be great to be able to direct link to a specific tag/list of tags too. e.g. ../creations/tags?="video,pickles"
> > 
> > 
> > Something similar is doable, I didn’t develop the feature to keep the PR small.
> 
> There's a PR with some of this functionality already submitted here by the way: #552 - it would need to work with properly with that format.

I saw that, I'd wait the merge of that plus this PR and branch from there ;)

esnho:
> > > Awesome. It would be great to be able to direct link to a specific tag/list of tags too. e.g. ../creations/tags?="video,pickles"
> > 
> > 
> > Something similar is doable, I didn’t develop the feature to keep the PR small.
> 
> There's a PR with some of this functionality already submitted here by the way: #552 - it would need to work with properly with that format.

Added tags to collection page and added parameters to share the page with some tags selected.

For example:
http://localhost:3000/tz/tz1Xye21UorssHqTHEyMQso5EG95sssea9vN?t=particles&t=feedback

Other example:
http://localhost:3000/tz/tz1Xye21UorssHqTHEyMQso5EG95sssea9vN/collection?t=code&t=3d

Market OBJKTs doesn't have any tags unfortunately, the code is very modular and can be ported to market section too if the market OBJKTs respect collection and creations section.

esnho:
Now is possible to filter per multiple tag by URL

image

crzypatchwork:
something similar could be used in a search feature but not in the profile in my opinion

esnho:
> something similar could be used in a search feature but not in the profile in my opinion

Why not? Is very useful to search through own collection and creations, can be used to share a tagged serie, or a subset of own collection.

for sure it’s useful in search too, but this was a low hanging fruit imho

andrevenancio:
Hi @esnho ! Thanks for your PR and comment.
Big visual changes like this normally go through a bit more scrutiny amongst a larger team including design and UX. We're already working on some iterations for this page but unfortunately the filtering wasn't prioritised. We decided to invest more on a Explore page which has access to the whole dataset of OBJKT we have on the platform. Though I understand that for users that have a huge amount of OBJKT in their possession a feature like this might be useful, we decided to implement pagination instead to address that issue.

Apologies if you're disappointed with this decision, maybe in the future we can review this decision.

esnho:
Hi @andrevenancio, @crzypatchwork,

I understand your decision, in a naive momentum I decided to implement this because I thought was helpful and was easy for me to implement without edit the backend, I’d preferred to have an earlier feedback but I know you are very busy.
I hope a way to create different personal galleries will be implemented or a better way to filter display pages. 
IMHO it would be very appreciated by those users with different styles in creations or collections.

I hope to find a better way to support HeN, I’m not able to focus on the hicaton in these days but I’ll keep an eyes on Trello in case I can find a ticket I can take and help if you agree.

Luca

andrevenancio:
As I mention, this type of sorting will be used on an explore page. currently when you click on a tag it takes you to hicetnunc.xyz/tag/XXXX it is in that page where this feature will be super useful. Happy if you wanna do a PR against that!

There's currently a team of 2 designers having a look at this and other parts of the site, so do drop by at Discord and get in touch with someone on the #dev channel. I'm not always there, but there's always someone there that could put you in the right direction!

Thanks Luca
Andre

PR #548, opened 2021-04-21
Artist icon and alias in feed and Objkt views

haustraliaer:
Currently you have to click right through to the wallet address page to see any info about the artist's profile (if they've set it up).

This will make it much more obvious if you're looking at the actual artist's work - either in the feed or on the objkt page.


andrevenancio:
This is visually a good feature. Though includes an extra request per objkt. Which on a few that loads 30item at a time, it will not do 30 extra requests.

Let me have a think about it also from the design point of view. There's a designer refactoring some parts of the site let me see what he thinks.

haustraliaer:
I tried to think of a way of avoiding the requests but came up with nothin’ that I could really do quickly on the front end.

I’m kind of assuming this api is the same as for objkts - no ability to send an array of addresses in one go, right?

If we set up a true global state machine for this kind of thing though, the initial request for any given wallet address (or objkt!) could easily be cached and then at least re-requesting the data would be avoided as you navigate around.

haustraliaer:
Also, related I guess - is there a way to bundle these profile requests in at the api level with objkts/wallet feeds etc?

you could take the same approach I’m talking about - cache on the server if it finds one so that all future requests for a given wallet are avoided?

haustraliaer:
@andrevenancio I've adjusted so it only makes the request on the Objkt view - not the feeds.

Slightly quicker understanding of who you're looking at (if they have a profile) seems better than not at all to me.

I still think we could pull profiles from a local cache (and avoid subsequent requests when loading wallet pages or other objkts from the same artist) - but I'd rather this get merged in first and I can add that as seperate PR, because we could probably do a bit more stuff with that concept as well.

PR #591, opened 2021-04-26
fix(): update beacon

AndreasGassmann:
Beacon SDK v2.2.5 contains a hotfix for an issue that prevents some users from connecting their P2P wallet (Kukai and AirGap).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@airgap/beacon-sdk": {
-      "version": "2.2.4",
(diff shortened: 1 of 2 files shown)


acolle:
Not sure if related but I'm still struggling to connect my Kukai wallet to H=N on Firefox (v 87.0). It works on Safari 14.0.3.

AndreasGassmann:
I don't think this change is deployed yet. Are you testing it locally?

This specific issue only exists if you have previously paired dApp <=> wallet.

acolle:
No not locally, I thought it had been deployed already, my bad.
It works in private windows so it must be a cache issue, ignore my previous message then

PR #595, opened 2021-04-27
Fix NaN on burn tab if don't own objkt

### LINKED ISSUE #592: Burn page should show zero instead of "NaN editions"
In some cases, 'Burn' page shows "NaN editions" instead of zero. This can be fixed with a default value in burn.js [here](https://github.com/hicetnunc2000/hicetnunc/blob/90affefb5f65d15400ecd693c884c0a7606a5bae/src/pages/objkt-display/tabs/burn.js#L14)

image

- OS: OSX 10.15.7 (19H524)
- Browser: Firefox 88.0

skenaja:
Fixes #592

### DIFF
--- a/src/pages/objkt-display/tabs/burn.js
+++ b/src/pages/objkt-display/tabs/burn.js
@@ -11,7 +11,7 @@ export const Burn = (props) => {
   const [amount, setAmount] = useState('')
   const [progress, setProgress] = useState() // eslint-disable-line
 
-  const totalOwned = parseInt(props.owners[address.address]) // check total the user owns of this token
+  const totalOwned = ( props.owners[address.address] && parseInt(props.owners[address.address]) ) || 0 // check total the user owns of this token
 
   const handleSubmit = () => {
     if (amount === '') {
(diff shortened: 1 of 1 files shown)


andrevenancio:
I'm not sure what to do with this PR... all of this is a bug because we're choosing not to fix another bug.
If you don't own an objkt you shouldn't see this tab... adding extra code just to play it safe is a bit of a hack in my opinion...

I think @crzypatchwork wanted to show the `swap` and `burn` for everyone and then rely on the fact that they don't own the objkt and the operation fails, instead of trying to get to the source of WHY is `owners` sometimes coming in without data. 

So I'm much more inclined on fixing that issue that adding extra code to the frontend just "in case it fails" somewhere else...

Thoughts?

skenaja:
You might be the creator of 100, with all 100 swapped. They're owned by the escrow account
You're allowed to view burn tab as creator.
As-is: you see the NaN error because you are owner of 0

Given they expect to be able to do burn as creator, they would probably be expecting to see the tab.
Hiding the tab if 0 owned might be a more confusing pattern.

skenaja:
9d41144 fixes #603

andrevenancio (approved):

PR #597, opened 2021-04-27
Filter banned wallets from objkt collectors tab.

neelts:
This will make life harder for the bots stealing and selling free OBJKT's


skenaja:
Possibly a better idea might be to remove the Collect button, and add a note to the wallet id eg `tz1... (banned)` to the owner-swap component.

That way it's more transparent to the creator where all their editions went, and also prevents users buying from banned users.

neelts:
> 
> 
> Possibly a better idea might be to remove the Collect button, and add a note to the wallet id eg `tz1... (banned)` to the owner-swap component.
> 
> That way it's more transparent to the creator where all their editions went, and also prevents users buying from banned users.

https://github.com/hicetnunc2000/hicetnunc/pull/598

andrevenancio:
This should probably be done on the API. I'm not sure adding this amount of logic to the frontend is a good idea. Can you do a PR on the API repo?

neelts:
> This should probably be done on the API. I'm not sure adding this amount of logic to the frontend is a good idea. Can you do a PR on the API repo?

I can, but have you checked this one [#598](https://github.com/hicetnunc2000/hicetnunc/pull/598)?
So which one would be better to do?

djangobits:
I prefer to leave a trace of banned objkts so the collector can decide to burn them.

PR #598, opened 2021-04-27
Remove collect button & add '(banned)' to the banned address.

neelts:
https://github.com/hicetnunc2000/hicetnunc/pull/597


neelts:
https://github.com/hicetnunc2000/hicetnunc/pull/597#issuecomment-827561275

djangobits:
that's cool! an other addition might be to create an overlay icon over the image on collection view if an objkt comes from a banned account so people would see very quick what stuff they have in their wallet. but let's see if this addition is merged first. I'd love it

PR #644, opened 2021-05-04
Cover/thumbnail upload and preview functionality

pichiste:
This PR adds a requirement on the mint page to upload a zip file containing cover/thumbnail media for every OBJKT that is minted. The user is required to use the external media compression tool (linked in the field description) to generate the zip file that is to be uploaded:
Screen Shot 2021-05-04 at 2 31 17 PM

After generating the zip file, uploading, and clicking "Preview", the cover/thumbnail media will also be displayed alongside the original OBJKT media:
Screen Shot 2021-05-04 at 2 32 11 PM


andrevenancio:
@pichiste such a good job mate! Get some rest and welcome the new family member! <3

pichiste:
Thanks! Please feel free to hit me up with any questions that may come up.

One note: Once the media compressor subdomain is sorted, you can set it here:
https://github.com/hicetnunc2000/hicetnunc/blob/feature/thumbnail_preview/src/pages/mint/index.js#L21

Issue #669, opened 2021-05-09
Feature: Increase max length of item description from 500 chars


skenaja:
**Is your feature request related to a problem? Please describe.**
The minting screen limits the item description to 500 characters. This is stored in JSON objkt metadata in IPFS. The TZIP-21 schema does not put a limit on `string`, and this should not affect any data stored in the blockchain.

**Describe the solution you'd like**
Increase the data to a more sensible 2000 characters.

PR #670, opened 2021-05-09
Increase maxlength of OBJKT description

skenaja:
Implements #669


andrevenancio (approved):

crzypatchwork:
maybe it could be even higher

skenaja:
I think if it goes much higher it should be in a `textarea` rather than `text` input.

Issue #662, opened 2021-05-08
Objkts with tags with % symbol cannot load


kylegrover:
**Describe the bug**
OBJKT#29859 cannot load, the UI crashes and just shows a grey screen with console errors:
`URIError: Pathname "/tags/96%biomass" could not be decoded. This is likely caused by an invalid percent-encoding.`

It must be the percent in the tag "96%BIOMASS"

**To Reproduce**
1. https://www.hicetnunc.xyz/objkt/29855

**Expected behavior**
Page will load briefly then go grey, console errors.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Brave
 - Version 1.24.82

kylegrover:
I assume the fix is something like uri encoding tag links as in /tags/96%25biomass

mtyka:
This can be closed.

PR #684, opened 2021-05-11
Bugs: filter was inverted and tags need encodeURI

mtyka:
1) The filter condition was inverted, it tries to filter out empty strings.
2) The tag is used as a URL and thus characters like '%' need to be URI encoded
Example that is currently broken in production: https://www.hicetnunc.xyz/objkt/29855
as the objkt contains a tag "96%biomass"
Noted already as: https://github.com/hicetnunc2000/hicetnunc/issues/662

### DIFF
--- a/src/components/tags/index.js
+++ b/src/components/tags/index.js
@@ -7,10 +7,10 @@ export const Tags = ({ tags }) => {
   return (
     <div className={styles.container}>
       {tags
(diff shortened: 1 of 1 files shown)


andrevenancio:
Thank you!

andrevenancio (approved):

PR #702, opened 2021-05-16
Add webxr template

michaelybecker:
adds a WebXR sample - a slight modification to the ThreeJS sample that enables XR rendering, accurate XR controller models and a standardized VR Button to get in/out of VR, per the best practices on the [ThreeJS Docs](https://threejs.org/docs/#manual/en/introduction/How-to-create-VR-content)


andrevenancio (approved):

andrevenancio:
thanks @MichaelHazani

PR #712, opened 2021-05-17
hDAO tooltip and layout changed on item-info component

violetforest:
- centered the loading circle
- tooltip on the hDAO circle on item-feed and objkt view
- item-feed component has some things switched around so it follows these formats in the pictures from the figma:
Screen Shot 2021-05-17 at 5 19 25 PM

and this one for objkt view (need to add the avatar later):
Screen Shot 2021-05-17 at 5 20 55 PM

_(the hdao stays on the left in the objkt view per rafa's request)_


andrevenancio (approved):

andrevenancio:
thanks 🇨🇴

PR #713, opened 2021-05-17
Telegram icon fix

360fun:
Telegram icon fix:

image
image


360fun:
it's been almost a month...when do you plan to deploy it? 🤔

PR #719, opened 2021-05-20
reverting back to model-viewer 1.6.0

andrevenancio:
Google's model-viewer has introduced breaking changes on version 1.7.0. They're updated to 1.7.1 to resolve some issues, but peple are still reporting issues. Reverting to 1.6.0 seems like the sensible thing to do.


andrevenancio:
This addresses https://github.com/hicetnunc2000/hicetnunc/issues/717

PR #709, opened 2021-05-17
Fix units of OBJKT's hdao balance and handle upstream API errors

skenaja:
Fixes #703

PR #722, opened 2021-05-20
hDAO collect - fix error on unsynced accounts

### LINKED ISSUE #703: hDAO reclaim amount display issues
**Describe the bug**
2 small issues with the new claim_hDAO feature on OBJKT page for creators, raised via Discord
https://discord.com/channels/759930487946215444/837237944208326667/843508469080391691

1. If the API doesn't return a value, the UI shows `-1`:

image

2 - The UI shows the value in microtez, not tez

image

Expected behaviour:

1. make the button behave as if the objkt has 0 hdao (so default behaviour)
2. divide the amount by 1,000,000 (as per user profile page)

skenaja:
Error reported: `TypeError: cannot read property 'address' of undefined`
Cause: unsync'd accounts have null `acc`.
Fix: test for null in check for collect or curate and tooltip

Reported in #709 
Fixes #703

### DIFF
--- a/src/components/item-info/index.js
+++ b/src/components/item-info/index.js
@@ -63,7 +63,7 @@ export const ItemInfo = ({
 
   const curateOrClaim = (id, balance = 0) => {
     // if user is creator and there's hDAO balance
-    if (acc.address === token_info.creators[0] && balance > 0) {
+    if (acc && acc.address === token_info.creators[0] && balance > 0) {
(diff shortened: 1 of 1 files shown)


andrevenancio (approved):

andrevenancio:
Thanks @skenaja ! This fixes it. I think I'm going to create some atom type components where they dont have any logic but just receive props. so we can add some tests against it.

PR #728, opened 2021-05-22
WIP -- Add twitter-card component

knowuh:
Issue: I don't think we can modify the meta tags after page load and still have twitter unfurl.  This is a problem because at page-load time we don't have an OBJKT ref.

### DIFF
--- a/src/pages/objkt-display/index.js
+++ b/src/pages/objkt-display/index.js
@@ -10,6 +10,7 @@ import { renderMediaType } from '../../components/media-types'
 import { ItemInfo } from '../../components/item-info'
 import { Menu } from '../../components/menu'
 import { Info, Collectors, Swap, Burn } from './tabs'
+import { TwitterCard } from './twitter-card'
 
 const TABS = [
   { title: 'info', component: Info }, // public tab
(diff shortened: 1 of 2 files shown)


andrevenancio:
You are correct. This is a problem the core team has been aware. It requires server side rendering. Some tools allow "react create app" to handle SSR but it requires the site to be hosted on a node.js server running express.io . At the moment the site is not hosted on a node.js server.

I've tried to convince the team to move the site to next.js but I think they want to prioritise other features. Maybe it's a good time to "pressure" him a little bit. (@crzypatchwork 😉)

Also, I know this is still work in progress but the "og" tag are tags for Facebook. You other need specific tags for Twitter.

knowuh:
@andrevenancio thank you for looking and confirming.  +1 for Next.js someday maybe with caching & proxy in front. Infrastructure costs should stay cheap/free if possible. 

Closing this.

andrevenancio:
Oh, you can reopen it and continue. I was only saying that in order for it to work there needs to be work done on the server. You need react helmet and you need to have the tags. But there are configurations in the server that need to happen for it to work.

PR #731, opened 2021-05-23
Hicathon: simple social sharing link

knowuh:
## Info:
We punted on doing static pages with OG Meta tags on the object page. (closed #728, can reopen later)

This PR is a pragmatic intermediate step, where we provide a sharing link that should unfold when used as social links.

Demo: https://hicathon2.netlify.app/objkt/94305  

- Look for the link icon to the right of the OBJKT ID# click to copy link to clipboard
- verify the link using https://cards-dev.twitter.com/validator

The Link button is here:
Screen Shot 2021-05-23 at 7 31 39 AM

The Social previews look like this:
Screen Shot 2021-05-23 at 8 22 21 AM

## Issues:
- Social link button relies on a third-party app, and there are probably concerns about such tight integration with a community tool. It uses http://objkt.link because its author was on our team (WG 2.3), though it would be easy to switch the provider.

- It's probably better to solve this problem in the long term using a more robust solution, TBD.


andrevenancio (commented):
I would consider putting a Boolean flag on the gallery file to enable this feature.

I can imagine that for The FEN specifically a random selection would not be the best solution. However for OBJKT4OBJKT this is a pretty interesting solution.

So I think a Boolean to activate this would be required IMHO

knowuh:
I think the above comment from Andre was meant for https://github.com/hicetnunc2000/hicetnunc/pull/732

crzypatchwork:
I would suggest using the following symbol? thoughts?

Screenshot from 2021-05-27 23-43-12

knowuh:
> I would suggest using the following symbol? thoughts?
> 
> Screenshot from 2021-05-27 23-43-12

I like @crzypatchwork's suggestion. That symbol seems more typical of 'copy' actions,  waiting for others to chime in.

kylegrover:
they're both fine imo. could push as is and do a dao vote for fun afterwards, or other way around.

PR #732, opened 2021-05-23
WG 7.1:  Add hicathon gallery

knowuh:
## What:

This PR comes from WG 7.1 to add a gallery for the #hicathon. 

Problem was that there were too many items to display at once.  This solution might not be ideal, but it works for now.

When we load a gallery that has more 100 items, we select 40 at random using Fisher-Yates shuffle.

03b4c95 -- adds the JSON for the galleries

b2f3e54 & ce3cba1 -- add the shuffle / limit logic.

## Live Demo

* https://hicathon-gallery.netlify.app/galleries
* https://hicathon-gallery.netlify.app/gallery/hicathon

## Documentation:

The gallery-details page now looks for a boolean flag in the `gallery-name.json` specifying `useRandomSample`. If that flag is present and true, the gallery page will only load up to  `maxSize` (if specified in the json, or default to 40) OBJKT to display.

## Additional TODO items:

It might be important to indicate that there are more OBJKT in the gallery than can be displayed at this time. Perhaps for now, we can ask the gallery curators to make this clear in the `description` field of the JSON.

Just opening this PR for the 7.1 WG team, and to start a discussion.


AliaK77:
@andrevenancio  @crzypatchwork  would you have a chance to review this please.

knowuh:
@andrevenancio  I think this comment somehow got in the wrong PR:

> I would consider putting a Boolean flag on the gallery file to enable this feature.
> I can imagine that for The FEN specifically a random selection would not be the best solution. However for OBJKT4OBJKT > this is a pretty interesting solution.
> So I think a Boolean to activate this would be required IMHO

Yes, a boolean flag in the JSON is so much cleaner.  Thank you.

Added that in 94f67a3

Thanks for having a look

knowuh:
Demo has been updated too: https://hicathon-gallery.netlify.app/

PR #744, opened 2021-05-24
mobile breakpoint padding on item display

violetforest:
10px padding instead of 20px on mobile to fix spacing issue, especially when objkts reach 100,000

hen1

FIX:
hen2


andrevenancio (commented):
Code looks fine, but screenshot seems like only the left has padding!

PR #751, opened 2021-05-24
Feature/UI fixes

violetforest:
Adding HTML tags to the objkt display page

Adding HTML5 tags for accessibility and SEO based on the following HTML5 guidelines:
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines

I'm open to suggestions on how to improve the hierarchy of h1,h2 tags or any other hierarchy, this is always super confusing o.o

Also wasn't sure what HTML tags to use for "TITLE" and "DESCRIPTION"


violetforest:
Adding HTML5 tags for accessibility and SEO based on the following HTML5 guidelines:
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines

I'm open to suggestions on how to improve the hierarchy of h1,h2 tags.

Also wasn't sure what HTML tags to use for "TITLE" and "DESCRIPTION"

src/components/item-info/index.js
```diff
@@ -96,57 +96,61 @@ export const ItemInfo = ({
...
-        </div>
-        {feed && (
-          <div className={styles.objktContainer}>
-            <Button to={`${PATH.OBJKT}/${token_id}`} disabled={isDetailView}>
-              <Primary>OBJKT#{token_id}</Primary>
-            </Button>
-            <div className={styles.hdaoButton}>
-              {renderHDAObutton(token_id, hDAO_balance)}
+        </footer>
+        <header className={styles.container}>
```
andrevenancio: Is this right? a `<footer>` before a `<header>` ? Do we need to be this strict with the markup?

andrevenancio:
Looks fine. just a small comment about semantics really.

src/components/item-info/index.js
violetforest: Heres a quote from the [MDN Guide](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines):

"HTML Footer Element `<footer>` defines a page footer, which typically contains copyright or legal notices and sometimes some links. In the context of a section, a footer might contain the sectioned content's publication date, license information, etc. `<article>, <section>, <aside>, and <nav>` can have their own `<footer>.` **Despite its name, the footer is not necessarily positioned at the end of a page or section.**"

violetforest: I can make it less strict if you guys want

violetforest: I took out the stricter tags in my latest commit

andrevenancio: I'm just not sure about the footer. As normally I would separate it from the article or section. in my head it makes sense to be a global thing like a header is, detatched from the current article being viewed. but its not the end of the world.

violetforest: i took a lot of html tags out / made everything simpler

PR #754, opened 2021-05-25
storybook: initial setup w/ token spec generation

### LINKED ISSUE #738: Storybook Support
**Is your feature request related to a problem? Please describe.**
WG 9.4 from the Hicathon will have some UX contributions to make in the form of a design system. While trying to follow the README to get setup to contribute, I've encountered some issues with packages not installing correctly, and also running the API. 

These are solvable issues, but ideally running the whole site wouldn't be required in order to work on the design system or make smaller UI contributions.

**Describe the solution you'd like**

Add Storybook to this repo:
1. enable faster UI development
2. allow contributions w/o a dependency on the `hicetnunc-api`
3. accommodate easier implementation & testing of the design components
4. enable easily spinning up coded prototypes that use the design system

I noticed this is mentioned in the Wiki as something already being explored. https://github.com/hicetnunc2000/hicetnunc/wiki/Contributing#testing

Has anybody started work yet? Or is this something I could begin in a PR.

**Describe alternatives you've considered**
1. Hand rolled design system folder (eg. plain HTML running a "design system" React app that loads components). Seems like unnecessary yak-shaving.
2. Design system managed via Storybook but in a separate repo. Seems to early to do this.

**Additional context**
Not familiar with React components and mocking state, fetched data, etc, but to get started I think we could break it down into phases:

+ **Phase 1: Design Tokens** eg. colors, fonts, spacing, breakpoints, etc
+ **Phase 2: Simple Components** anything that doesn't require being connected to state or having us mock things (eg. `getLanguage` or anything else)
+ **Phase 3: Mocked Components** anything that requires more ceremony to display correctly and isolate from whatever app context it expects
+ **Phase 4: Test Harness** to add coverage to key components and protect against basic regressions

depatchedmode:
Fixes #738 

So far this just adds Storybook with the Design Token add-on and quickly instruments the `variables.scss` file for an example. 

Next on the list:

- [x] add an example story for the Button component
- [x] await feedback on new [shade palette and type scale proposal](https://www.figma.com/file/Q2UWiIm6FOOBmscXrMVid6/WG-9.4?node-id=657%3A2). This has now been accepted into the post-hicathon [UI/UX WG designs](https://www.figma.com/file/R5sw1HcoHx4Aup1Yl7rGYA/WG-UI-UX?node-id=1261%3A11247): 
- [ ] add these new tokens to the spec
~~- [ ] audit and document any other css custom properties / scss variables that should be considered core design tokens~~
~~- [ ] breakpoints audit~~
~~- [ ] basic layout grid~~


andrevenancio:
So I'm all in to bring Storybook into the project. But..
The design done during hickaton is done by the community. Pretty good stuff in there, however, for the last couple of week we have a dedicated team of designers that have been working on a new version of the design.

I'm not the one who should be deciding what design needs to be implemented. Its my understanding that Raphael (hickaton) has been in touch with Julio (Designer) but I'm not sure how they're going to move forward.

I would not make major changes in new components until that has been agreed.

@crzypatchwork would you mind giving Julio and Raphael some direction of what would you like to keep from the 2 different designes we have at the moment?

package.json
```diff
@@ -49,10 +49,22 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "eb:prod": "npm run build && npm run start:prod",
-    "prettier": "prettier --write \"./src/**/*.{js,jsx}\""
+    "prettier": "prettier --write \"./src/**/*.{js,jsx}\"",
+    "storybook": "start-storybook -p 6006 -s public",
```
andrevenancio: Can you remove `public` ? we dont want it to build to the public folder but rather for Storybook folder `storybook-static`. This will allow us to coordinate 2 different deployments. The public folder should only be for the react-create-app application

src/stories/tokens-animation.stories.mdx
```diff
@@ -0,0 +1,6 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+import { DesignTokenDocBlock } from 'storybook-design-token/dist/doc-blocks';
+
+<Meta title="Design Tokens/Animation" />
```
andrevenancio, line 4: Would be nice to also use the `controls` feature from Storybook

depatchedmode:
@andrevenancio Fully agree. Wasn't intending for this PR to deal w components at all, yet. Just to formalize design tokens, and that first to do is just noting that there are some proposals that need to be agreed upon, pending feedback from the core team. 

I saw that you raised concerns in the Figma about contrast and I agree. That's why I put together that "shade palette" proposal with WCAG AA/AAA contrast ratios details for both the light and dark mode themes. If the core team has a different one, also cool.

Just want to ensure that fundamental palette of shades and type sizes is nailed for everyone in branch. Feels like that bit is achievable and is the best first Story.

depatchedmode:
@andrevenancio oh one other question I did have:

there are both scss variables and css custom properties in the codebase. it looks like custom properties is the preferred approach, as that's used in the components. are you moving away from scss or is the intention to use both?

andrevenancio:
> @andrevenancio oh one other question I did have:
> 
> there are both scss variables and css custom properties in the codebase. it looks like custom properties is the preferred approach, as that's used in the components. are you moving away from scss or is the intention to use both?

So the project started with css. I moved in and implemented sass, there might be some old css laying around.

We've been talking about moving this project to next.js. So that's something to consider too.

src/stories/tokens-animation.stories.mdx
depatchedmode: I believe those are included by default and bind automatically where supported now:
https://storybook.js.org/docs/react/essentials/controls

depatchedmode: Can still be configured custom, but probably a pattern to test when we get to the components phase.

crzypatchwork:
as I read these days on twitter cypherpunks code...

package.json
```diff
@@ -49,10 +49,22 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "eb:prod": "npm run build && npm run start:prod",
-    "prettier": "prettier --write \"./src/**/*.{js,jsx}\""
+    "prettier": "prettier --write \"./src/**/*.{js,jsx}\"",
+    "storybook": "start-storybook -p 6006 -s storybook-static",
```
andrevenancio: this is fine too, but if you just add  `start-storybook -p 6006` it will default to `storybook-static`.

src/stories/tokens-animation.stories.mdx
andrevenancio: What do you think of this pattern?
```
import { Story, Meta } from '@storybook/react/types-6-0';

import { Example, ExampleProps } from './index';

export default {
  title: 'Example',
  component: Example,
  args: {
    children: 'An example',
  },
} as Meta;

const Template: Story<ExampleProps> = (args: any) => {
  return <Example {...args} />;
};

export const Basic = Template.bind({});

```

andrevenancio (approved):
all looks fine. just some minor comments.

I would probably focus on the src/components/layout

src/stories/tokens-animation.stories.mdx
depatchedmode: No complaints here.

depatchedmode: ok - wasn't seeing much movement yet on the Figma design system proposals, so thought it might just be good to add an example story for the Button component.

I don't have a ton of recent React experience, but I followed the pattern above as best I could minus the Typescript aspects which I don't think are established in this repo yet.

Let me know if you want anything tweaked.

crzypatchwork:
this was an interesting thing for documenting the protocol if I understood it right

depatchedmode:
> this was an interesting thing for documenting the protocol if I understood it right

Yeah, the MDX files would be great for that in addition to documenting the design system.

miguelzinhe:
> > this was an interesting thing for documenting the protocol if I understood it right
> 
> Yeah, the MDX files would be great for that in addition to documenting the design system.

I can help with this addition!

depatchedmode:
I've rebased onto `main`, and resolved the conflicts. Added a bit back to the scope of this, as it seems like it's worth getting the design tokens into this as a part of the post-hicathon UI/UX working group's efforts.

If these simple design tokens are accepted into main, it will make the other PR's (nav, minting page, etc) easier to implement and more consistent with each other.

Design tokens will be:
+ shade palette (light & dark mode)
+ type sizes
+ spacing units

PR #762, opened 2021-05-26
Added sort selection control to artist page for creations/collection

dialectic-design:
One of the outcomes of hicathon WG1.1 working on sorting/categories.

Added a new component 'simple-select' which adopts the hicetnunc theming (dark/light/fonts etc).
Used this component on the 'display' page in a secondary menu (below the creations/collection tabs) for the user to select a sorting option for the objkts displayed [newest, oldest, random].

A demo of this control is running here:
[https://hicetnunc-demo-test-vagak.ondigitalocean.app/tz/tz1QRK6K5LyEjdPCyeV6aSeBQ8z6Tk5musz5](https://hicetnunc-demo-test-vagak.ondigitalocean.app/tz/tz1QRK6K5LyEjdPCyeV6aSeBQ8z6Tk5musz5)

Please let me know if there is anything you'd like to see changed before merging!.

Or if this is a feature you are not interested in adding right now. Then I can redirect my attention elsewere :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -165,9 +165,9 @@
       }
     },
     "node_modules/@babel/helper-define-polyfill-provider": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/@babel/helper-define-polyfill-provider/-/helper-define-polyfill-provider-0.2.0.tgz",
-      "integrity": "sha512-JT8tHuFjKBo8NnaUbblz7mIu1nnvUDiHVjXXkulZULyidvo/7P6TY7+YqpV37IfF+KUFxmlK04elKtGKXaiVgw==",
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/@babel/helper-define-polyfill-provider/-/helper-define-polyfill-provider-0.2.1.tgz",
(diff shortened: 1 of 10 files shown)


andrevenancio:
Hi @dialectic-design Thanks for submitting your code. I had a chance to have a look at what you're proposing, and the demo and I have some notes.

As rule of thumb we try to avoid installing new dependencies unless strictly necessary. What you need is pretty simple and could be done in 5 lines of HTML
```jsx
<select>
  <option value="newest">newest</option>
  <option value="oldest">oldest</option>
  <option value="random">random</option>
</select>
```
If you want to make it a component, you could do something like [this](https://github.com/hicetnunc2000/hicetnunc/blob/main/src/components/button-language/index.js). Simply because the component you created is 285 lines of code and has 3 css and 3 scss files.

So, simplify it a bit.

The sorting code on display/index.js is alright. My only other note is to avoid adding the negative margin-top there, so maybe you can refactor the html there too so fit inside the <Containar> that holds the creations/collection.

dialectic-design:
Thanks for the feedback @andrevenancio and understood!

I pulled in code for a React Select component with more features (multi-select etc) but indeed it would add code currently un-used. 

Following your link to the language select 'button' should I rename this to sort-select for example?

andrevenancio:
> Following your link to the language select 'button' should I rename this to sort-select for example?

Hey @dialectic-design ! 

- [ ] Please rename it to select.

- [ ] Make sure there's nothing hardcoded on that component, but it receives an array for the options instead

- [ ] Remove any references you had from css files and lets get this PR slimmer!

PR #772, opened 2021-05-27
fix #767

### LINKED ISSUE #767: inputs missing "name" attribute / no autofill
**Describe the bug**
On the minting page none of the input fields have a 'name' attribute set, resulting in them all defaulting to `name="input-name-not-set"`. As a result of this, the text fields do not remember your previous entries. This is bothersome to users when their mint fails and they lose their description, title, and tags.

**To Reproduce**
Mint an objkt. Return to the mint page and click on an input field. Your browser should show you your previous entries as autofill options, but instead it will show the previous 'title' entries, no matter which input you click on.

**Expected behavior**
Clicking on an input field should show your previous entries as autofill options.

 - OS: Windows 10
 - Browser Brave
 - Version latest

**Additional context**
[I believe this is the relevant code](https://github.com/hicetnunc2000/hicetnunc/blob/d6fde0bfd6eb37fe01b8497f7bcddff6a41442bc/src/pages/mint/index.js#L257)

**QUESTION:** Is there any reason the fields do not have `name` attributes? Will the form processing logic be affected if I add sensible name attributes to these input elements?

kylegrover:
**note: not tested**

I don't have a full test environment for this, so please do ensure these name fields don't throw a wrench in the form processing. It looks like the data is all handled through onChange anyway, but I mention it just to be cautious.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7755,9 +7755,9 @@
       }
     },
     "fflate": {
-      "version": "0.6.10",
-      "resolved": "https://registry.npmjs.org/fflate/-/fflate-0.6.10.tgz",
-      "integrity": "sha512-IQrh3lEPM93wVCEczc9SaAOvkmcoQn/G8Bo1e8ZPlY3X3bnAxWaBdvTdvM1hP62iZp0BXWDy4vTAy4fF0+Dlpg=="
+      "version": "0.7.0",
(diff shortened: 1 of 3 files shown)


crzypatchwork:
I feel one thing that should be added in there is that the description field should have a flexible size. Don't know if this considers it...

PR #786, opened 2021-05-29
reverse title to "{title} - hic et nunc"

kylegrover:
proposal to switch from: "hic et nunc - {title}" to "{title} - hic et nunc" to make browser tabs easier to navigate.
Fixes this problem when you have lots of tabs:
image
After:
image


andrevenancio (approved):
looks good!

PR #788, opened 2021-05-30
clickable pdfs

### LINKED ISSUE #763: Cannot navigate to a PDF from a collection page
**Describe the bug**

Since the recent changes that removed the ID overlay from each on a collection page, you can't click on a PDF OBJKT and navigate to its own page.

**To Reproduce**
1. Go to https://www.hicetnunc.xyz/tz/tz1LfkiegSFauPXzh4pwsAKC2ddqHgNKy89J
2. You can interact with the PDF but you cannot reach the OBJKT page

**Expected behavior**
You should be able to navigate to https://www.hicetnunc.xyz/objkt/20147

 - OS: Ubuntu 1804
 - Browser: Firefox
 - Version: 88

kylegrover:
fix #763 by setting 'pointer-events: none' on pdf viewer in feeds. 

this creates a somewhat unnatural UI where you can't scroll the pdf viewer until you go to the objkt page, but fixes the current huge issue of objkts being inaccessible through the UI. long term solution is a preview image.

### DIFF
--- a/src/pages/display/index.js
+++ b/src/pages/display/index.js
@@ -445,6 +445,7 @@ export default class Display extends Component {
                         mimeType,
                         uri: uri.split('//')[1],
                         metadata: nft,
+                        pdfscroll: false
                       })}
                     </div>
                   </Button>
@@ -469,6 +470,7 @@ export default class Display extends Component {
                         mimeType,
                         uri: uri.split('//')[1],
                         metadata: nft,
+                        pdfscroll: false
                       })}
                     </div>
                   </Button>
--- a/src/components/media-types/index.js
+++ b/src/components/media-types/index.js
@@ -28,6 +28,7 @@ export const renderMediaType = (props) => {
     interactive = false,
     preview = false,
     metadata,
+    pdfscroll = true
   } = props
   const path = uri
   let url = preview ? uri : `${CLOUDFLARE}${path}`
@@ -118,7 +119,7 @@ export const renderMediaType = (props) => {
     case MIMETYPE.PDF:
       return (
         <Container interactive={interactive}>
-          <PdfComponent src={url} />
+          <PdfComponent src={url} pdfscroll={pdfscroll} />
         </Container>
       )
     default:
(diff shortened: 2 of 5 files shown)


src/pages/display/index.js
```diff
@@ -445,6 +445,7 @@ export default class Display extends Component {
                         mimeType,
                         uri: uri.split('//')[1],
                         metadata: nft,
+                        pdfscroll: false
```
marualderete, line 448: Not a review, just want to understand: 
I dont see the connection between the changes in PdfComponent and this. Is this a pdfComponent too? just confused because I see `src/components/media-types/index.js` , where it is clear that there is a pdfComponent implementation there, but not here. In any case, if this component is being used in other places where the content it is NOT a pdf, maybe a more generic name would be better, or a hierarchy with a new component that should be used only for pdf elements would be more cleaner and better to maintain.

andrevenancio (commented):
I would avoid creating very specific props that need to be drilled in. Though I understand why you added the pdfscroll, I wonder if we could use `pdfscroll={interactive}` Just because interactive is set to `true` once you're on the OBJKT detail page, and its on that point that I think you would want to allow people to scroll through the PDF?

kylegrover:
Sounds right to me. I tried to look for an existing prop to use but didn't find that. The intention is to have this apply everywhere except the objkt detail page. I can rewrite with that prop.

kylegrover:
one consideration for the prop - I think it should be true on the objkt display page and also the home feed page, because the home feed page displays large and has an objkt link below it. It only needs to be false where there is no objkt link. This issue/pr is also moot if the OBJKT# links are returned to the creation/collection/tag pages, or if we provide some other way to link to PDFs

kylegrover:
Going to rewrite prop and/or use react-pdf when I get to it and resubmit

crzypatchwork:
I would suggest to add clickable for mp3/pdf which are unabled right now as a hover. when you hover it shows up the objkt number with a link

PR #790, opened 2021-05-30
Add Hicdex API to the CSP

aevin-io:
Requesting a PR to add **[](https://api.hicdex.com/)** in the Allowed Domains list ( CSP ) in order to allow Interactive Objects to query the Hicdex.com GraphiQL API. 
File to be modified ( under the 'connect-src' section - lines 166-173)
[](https://github.com/hicetnunc2000/hicetnunc/blob/5bd98a851827e3ab979bc29351e5ee8682a47450/src/utils/html.js)


aevin-io:
Please ignore

PR #799, opened 2021-06-01
video loading=lazy

kylegrover:
adds `loading="lazy"` to video render element. [Browser support 70%](https://caniuse.com/loading-lazy-attr), fallback behavior for other browsers is the current behavior.
In testing on a large collection this sped up the above-the-fold load time noticeably, and reduced initial page load by half a gigabyte. 
image_614

Tested on windows chrome/brave, might be worth browser testing.


sf-pixel:
Works with windows chrome, firefox (both dev and regular), and edge. Good stuff.

PR #813, opened 2021-06-02
Hotfix for displayUri bug in interactive objkt display

pichiste:
The `displayUri` value was being incorrectly set using `thumbnail_uri`:

Screen Shot 2021-06-02 at 7 24 47 AM

Have updated to use 'display_uri` and now images display properly:

Screen Shot 2021-06-02 at 7 24 29 AM


pichiste:
@crzypatchwork @andrevenancio going ahead and merging this to main

PR #820, opened 2021-06-02
Add services.tzkt.io to the CSP

aevin-io:
Requesting PR to include https://services.tzkt.io in the CSP, under the 'img-src/blob' definition
to allow Interactive OBJKTs to load images such as users avatars stored on site.


andrevenancio:
Looks good!

PR #861, opened 2021-06-10
implement react-pdf

### LINKED ISSUE #763: Cannot navigate to a PDF from a collection page

kylegrover:
Probably some untested devices / scenarios but I think this is a solid start.
image
image

Potential improvements: 
- align 'Loading PDF/Page' text
- better 'fullscreen' display
- nav arrows
- limit height/crop in feeds

fixes #763

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13376,6 +13376,11 @@
         "sourcemap-codec": "^1.4.4"
       }
     },
+    "make-cancellable-promise": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/make-cancellable-promise/-/make-cancellable-promise-1.0.0.tgz",
+      "integrity": "sha512-+YO6Grg2uy/z8Mv3uV90OP6yAUHIF43YGgEFbejmBrK9VWFsVO6DvzFMcopXr9wCNg3/QIltIKiSCROC7zFB2g=="
+    },
     "make-dir": {
       "version": "2.1.0",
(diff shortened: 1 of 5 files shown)


andrevenancio (approved):
Looks fine. but a couple of days ago @veqtor submitted a fix for an PDF Exploit. do you know if this new react-pdf protects against it? Basically javascript could be injected through a PDF.

In terms of style, I would just use the current buttons we have instead of <button>.

So something like 
```jsx
<Button>
  <Primary>blabla</Primary>
</Button>
```

maybe?

kylegrover:
Fixed fullscreen display, added annotation support, and switched to `<Button><Primary>`
image
I think this probably needs more testing with various PDFs... but PDFs are currently pretty broken so maybe we can just push this then sort it out later.

I tested on Nahiko's /objkt/124927 - it fails to load entirely as it's not a PDF at all. I think this is acceptable.
image

PR #882, opened 2021-06-14
Update w.json

markknol:
missing comma


kylegrover:
note: the obvious fix here is to not make this mistake, which I won't because I'll be hyper-paranoid about checking for commas now... but a more durable solution would be to check this file before it's committed, or otherwise create a system where typos are impossible.

PR #899, opened 2021-06-16
Add a popup component

sf-pixel:
### Simple Popups
---
An example of the component (commented out of course) can be found in the about.js file. The only parameters are sad, happy, exclamation, and null. Each corresponds to the respective svg, when null or not sad, happy nor exclamation, the HEN svg is returned.

Feedback please :). (First PR)


sf-pixel:
**Forgot to mention there's a title parameter, simply returns the title of the popup as a bolded text

src/components/popup/index.js
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react'
+import styles from './styles.module.scss'
+
+export const Popup = ({
+    sad,
+    happy,
+    exclamation,
+    question,
```
kylegrover: looks like 'question' isn't used anywhere, an abandoned variable maybe?

kylegrover:
some screenshots for reference. I started reviewing this and think it looks pretty good code-wise with maybe some aesthetic feedback, will give it some more thought later or if anyone else has feedback that's great.
image_657
image_658

couple of ideas: white border instead of box shadow, hacky backdrop with `box-shadow: 0 0 0 99999px #1119;`, and a bigger close button (tiny close buttons are no fun on mobile)
image

src/components/popup/index.js
sf-pixel: Yep, didn't catch it... figured that a question mark wasn't something needed, forgot to clean it up

kylegrover:
@Pixel10312 can you update to remove the demo? needs to be out before merge into live site. also let me know what you think on the CSS ideas.

sf-pixel:
> @Pixel10312 can you update to remove the demo? needs to be out before merge into live site. also let me know what you think on the CSS ideas.

Going to commit the CSS changes, really liked the 'focused' view. Increased the `close__btn` font size to `x-large`, should make a difference when on mobile. Will also remove the demo.

PR #902, opened 2021-06-17
OBJKT Width display on High-density pixel screens

aevin-io:
Requesting a PR to allow a unified display of objects across various screen resolutions; while keeping as much information as possible visible underneath the object preview.
Using the a 'vw' derivative for the width, creates a similar layout // look-n-feel, if viewed on either conventional pixel density displays or retina. Currently the object preview appears rather small on retina displays, and it mostly affecting video and static image artworks which are horizontal and specially interactive objects which require the user to navigate/use the artwork. Thank you.

Current display, on retina
image

Retina using 36vw
image

1920x1200 , using 36vw
image

The last two are very similar, although viewed in much different resolutions.

### DIFF
--- a/src/pages/objkt-display/styles.module.scss
+++ b/src/pages/objkt-display/styles.module.scss
@@ -40,7 +40,7 @@
   align-items: center;
 
   @include respond-to('desktop') {
-    width: 60vh;
+    width: 38vw;
   }
 
   > div {


kylegrover:
I took my own stab at this with #908 , I think it solves this problem

andrevenancio:
I also was having a look at this yesterday.. I got it as big as I "legally" could :)

PR #908, opened 2021-06-18
Objkt display height

kylegrover:
increases objkt display page height for images and videos, shrinks & evens the margin around objkt-display container, remove box-shadow around .info, and fix broken galleries


kylegrover:
image

andrevenancio:
Hey! Yesterday I've tried to raise that to 100vh and showed it to Rafael and he wasn't a fan. I managed to "push it" so it goes from 60vh to 70vh but check with him.

llumidev:
> 
> 
> Hey! Yesterday I've tried to raise that to 100vh and showed it to Rafael and he wasn't a fan. I managed to "push it" so it goes from 60vh to 70vh but check with him.

I think this kind of decision should be based on some community vote or something like that, not on someone's personal approval, most of the artists I know disapprove of the change in size of the objkts (which used to be bigger)

aevin-io:
> > Hey! Yesterday I've tried to raise that to 100vh and showed it to Rafael and he wasn't a fan. I managed to "push it" so it goes from 60vh to 70vh but check with him.
> 
> I think this kind of decision should be based on some community vote or something like that, not on someone's personal approval, most of the artists I know disapprove of the change in size of the objkts (which used to be bigger)

i couldn't agree more.

crzypatchwork:
@llumidev participation > vote

kylegrover:
I think there's some healthy balance of all of this stuff. Participation is a must or nothing happens, but we can sometimes use voting to guide our changes so we can breeze past these kinds of discussions. Also think this should be a healthy height balance, I used calc to set it so the collect button is always just barely in view, it ends up being pretty close to 70vh I think

kylegrover:
One minor change should maybe be made: I think I hardcoded `#222` as the border color, should maybe be transparent gray or set by a css var so it's lighter on light mode. Sometimes I forget there are people out there not using dark mode 😎. 
Also I only have a couple of devices so if anyone can test on Mac, Android, anything else I might forget that's always beneficial for these changes.

Karl-010:
@andrevenancio could you try to talk to Rafael? To at least reach 90vh, it was much better before, I work with interactive pieces and this smaller size is so bad (I'm not the only one who thinks so), not only for interactive works but all the other types. Pleaseeee🙏

kylegrover:
I added a proposal in basedao to set `hicetnunc2000/hicetnunc/pull/908` to `merge` so we can try voting on this. This will be non-binding since we haven't declared that as a mechanism, but I think it's worth an experiment.

Anybody have recommendations for creating a public demo? Just deploy to Netlify / other static hosting? Would be nice to allow non-developers to try out PRs and vote / give feedback.

bilgekaan:
> @andrevenancio could you try to talk to Rafael? To at least reach 90vh, it was much better before, I work with interactive pieces and this smaller size is so bad (I'm not the only one who thinks so), not only for interactive works but all the other types. Pleaseeee🙏

same, my interactive pieces are looking really bad right now.

for example: https://www.hicetnunc.xyz/objkt/141104

the iframe was around 600px height when I mint this and it was twice as big because its original size is 300x300 and it uses letterbox integer scale.

bilgekaan:
now the iframe is too big for interactive objkts

PR #913, opened 2021-06-18
Develop

crzypatchwork:


andrevenancio:
This will likely break develop like it did a couple of days ago when the minting got broken.

Basically @pichiste has a fancy version on develop that is better than the version on main for the mint. (The resize of the videos, and images for displayUri).

But you never told me to merge it @crzypatchwork so I never did. But basically it asks users to go to an external site to generate thumbnails, it allows the users to download the thumbnails and then they need to upload the zip file they got from that external site.

That version is better, but its still only working on develop.

Would be nice if you could test it, and let us know if you want it on main. I can merge it after. Its quite different on the `mint` and on the `ipfs.js` file that mint calls

PR #916, opened 2021-06-18
base32 encode cidv1 subdomain

kylegrover:
var subdomain = cidv1.toString() -> var subdomain = cidv1.toBaseEncodedString('base32')
seems to fix


andrevenancio (approved):
merging to test.

PR #917, opened 2021-06-18
dont show your objkts in feed, message if you haven't collected anything

violetforest:
- dont show your own objkts in the feed
- show a message if you haven't collected anything


andrevenancio (approved):
When are we launching this? Looks good!

PR #927, opened 2021-06-20
Config subjkt validation

kylegrover:
setting to draft because the subjkt_url field isn't set up to actually do anything currently, can merge after setting that up and testing or just use some of the other design changes and nix the subjkt_url idea


kylegrover:
screens
image
image

PR #954, opened 2021-06-27
Decoded SUBJKTS in the URL

sf-pixel:
Spaces, special characters and any otherwise broken profile links are now decoded before sending SUBJKT query. Decided to do the PR against the main as I've added less than 20 characters.

### DIFF
--- a/src/pages/display/index.js
+++ b/src/pages/display/index.js
@@ -251,7 +251,8 @@ export default class Display extends Component {
 
       this.onReady()
     } else {
-      let res = await fetchSubjkts(window.location.pathname.split('/')[1])
+      let res = await fetchSubjkts(decodeURI(window.location.pathname.split('/')[1]))
+      // console.log(decodeURI(window.location.pathname.split('/')[1]))
       console.log(res)
 
       this.setState({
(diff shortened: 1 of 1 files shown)


kylegrover:
Can you give an example of a broken subjkt that this fixes so I can test it? Also it seems to break if you feed it an invalid subjkt, might need some error handling.

image

sf-pixel:
> Can you give an example of a broken subjkt that this fixes so I can test it? Also it seems to break if you feed it an invalid subjkt, might need some error handling.
> 
> image
> 

Didn't see that coming, will for sure be adding validation into it, wasn't completely sure on broken accounts that are still broken (since a lot of people changed their subjkts to omit spaces recently) but query 30 or so subjkts on hicdex and you'll have one with spaces in between (currently displayed as: name%20ending) where the subjkt is properly encoded but never decoded.

kylegrover:
`/Yuvraj Imaginaria` is an example I found, seems to work! I think if you get the error handling fixed this should be merged.

sf-pixel:
@kylegrover invalid SUBJKTs now redirect to the feed.

PR #960, opened 2021-06-29
Temporary alert banner

kylegrover:
quickly adding a bottom-banner component to alert people about the swap/collect situation, on objkt pages only


kylegrover:
image_668
image_669

PR #983, opened 2021-07-07
Change default storage to prevent backtracks

sf-pixel:
Default storage was not set within the v2 swaps function, adding ```{ storageLimit: 209 }``` to the ```.send()``` function sets the default storage to 209 (the most storage that has been used with swaps AFAIK) instead of the default.

### DIFF
--- a/src/context/HicetnuncContext.js
+++ b/src/context/HicetnuncContext.js
@@ -104,7 +104,7 @@ class HicetnuncContextProviderClass extends Component {
         let batch = await Tezos.wallet.batch().withContractCall(objkts.methods.update_operators([{ add_operator: { operator: this.state.v2, token_id: parseFloat(objkt_id), owner: from } }]))
         .withContractCall(marketplace.methods.swap(creator, parseFloat(objkt_amount), parseFloat(objkt_id), parseFloat(royalties), parseFloat(xtz_per_objkt)))
 
-        return await batch.send()
+        return await batch.send({ storageLimit: 209 })
       },
 
       batch_cancel: async (arr) => {
--- a/src/context/HicetnuncContext.js
+++ b/src/context/HicetnuncContext.js
@@ -104,7 +104,7 @@ class HicetnuncContextProviderClass extends Component {
         let batch = await Tezos.wallet.batch().withContractCall(objkts.methods.update_operators([{ add_operator: { operator: this.state.v2, token_id: parseFloat(objkt_id), owner: from } }]))
         .withContractCall(marketplace.methods.swap(creator, parseFloat(objkt_amount), parseFloat(objkt_id), parseFloat(royalties), parseFloat(xtz_per_objkt)))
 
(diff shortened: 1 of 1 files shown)


BitPsycho:
The storage fee field should just be editable imo.  Before I always had put it to 400 as below that I still occasionaly had backtracking issues.

sf-pixel:
> The storage fee field should just be editable imo.  Before I always had put it to 400 as below that I still occasionaly had backtracking issues.

It is editable, this just prevents the default from being _below_ 209

skenaja:
I don't think that Temple allows editable storage (yet?) on batch tx (a v2 swap is an [update_operator + swap tx](https://better-call.dev/mainnet/opg/opTWqA9bW1TsfmuVdK6tsRXqUHXFrkNpHVvsTPXJN9HpDJKeuqV/contents))

sf-pixel:
> I don't think that Temple allows editable storage (yet?) on batch tx (a v2 swap is an [update_operator + swap tx](https://better-call.dev/mainnet/opg/opTWqA9bW1TsfmuVdK6tsRXqUHXFrkNpHVvsTPXJN9HpDJKeuqV/contents))

ah, wasn't aware that temple didn't allow that, I mainly use Kukai so that's why... that said, since storage is only consumed on an as-needed basis, how much higher should I increase the storageLimit to ensure most if not all swaps go thru? I honestly don't know since I do swaps thru a script with 1000 as my limit.

crzypatchwork:
I've tried a similar solution but I was not able to see it picking up in the raw transaction. A similar solution was presented by nft biker also which was recently uploaded.

BitPsycho:
> I don't think that Temple allows editable storage (yet?) on batch tx (a v2 swap is an [update_operator + swap tx](https://better-call.dev/mainnet/opg/opTWqA9bW1TsfmuVdK6tsRXqUHXFrkNpHVvsTPXJN9HpDJKeuqV/contents))

Ah wasn't aware of that .. if that is the case then they should just default to a high number no ?  If I understood this whole storage fee then only what is really used is payed for anyways ?  ( On a side note - anyone have a good link that describes the storage feel concept on the Tezos blockchain and how it relates to gas fees ? )

BitPsycho:
> > I don't think that Temple allows editable storage (yet?) on batch tx (a v2 swap is an [update_operator + swap tx](https://better-call.dev/mainnet/opg/opTWqA9bW1TsfmuVdK6tsRXqUHXFrkNpHVvsTPXJN9HpDJKeuqV/contents))
> 
> ah, wasn't aware that temple didn't allow that, I mainly use Kukai so that's why... that said, since storage is only consumed on an as-needed basis, how much higher should I increase the storageLimit to ensure most if not all swaps go thru? I honestly don't know since I do swaps thru a script with 1000 as my limit.

Yeah been using 350 before v2 and that always worked fine.

sf-pixel:
PR is closed thanks to a solution from @nftbiker.

PR #987, opened 2021-07-08
Feature/opengraph previews

csb324:
Hi! This may be wildly off-protocol, but I thought I'd try to add the display uri to the objkt page metadata. I'm realizing this might need to be server-side, but, worth a try.

Also noticed the Page component was duplicating 'hic et nunc' so I switched the default to be an empty string, which the component seems to expect. 

Anyway, hope this helps us share our art :)


csb324:
Closing this because for it to work it would have to be server-side, and that's a whole other can of worms. The title fix is over here -- https://github.com/hicetnunc2000/hicetnunc/pull/988

PR #993, opened 2021-07-10
swaps tab (for active non-v1 swaps)

kylegrover:
view active swaps from a wallet. query might want some refining, everything works fine on my account at least.


kylegrover:
image_674

kylegrover:
covered by violet's recent updates

PR #994, opened 2021-07-10
Feature/header profile action

jonnyscholes:
I regularly click the wallet preview expecting to goto my profile, but nothing happens. Hovering makes it look like a link, so it feels odd that nothing happens.

This PR links the wallet preview to the profile page of the synced wallet. I have decoupled the unsync from the profile button. Instead it now exists as a separate button to the left of the wallet preview when the menu is opened.

Removed some CSS that didnt seem to be used anywhere whilst I was here (`.hamburger` code).

Screen Shot 2021-07-10 at 11 25 10 pm

### DIFF
--- a/src/components/header/index.js
+++ b/src/components/header/index.js
@@ -37,27 +37,21 @@ export const Header = () => {
   }
 
   // we assume user isn't connected
-  let profileButton = (
-    <Button onClick={()=>context.syncTaquito()} secondary>
-      <Primary>sync</Primary>
-    </Button>
-  )
+  let headerButtonHandler = () => context.syncTaquito()
+  let headerButtonText = 'sync'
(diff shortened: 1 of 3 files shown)


kylegrover:
had the same thought recently! glad someone decided to do something about it. 
I'm not sure if the decoupling part is an improvement though, and it might even cause problems on mobile. but definitely yes to the linking, and yes to the code cleanup

jonnyscholes:
Sure, it worked ok down to 375px, but your right below that it got too cramped.

I've reverted it to be much closer to what it was, but with the wallet preview still being a link 🙂

jonnyscholes:
Closing this in favour of #1007 which covers this off :)

PR #1003, opened 2021-07-17
fix swap links

kylegrover:
problem: https://hicetnunc.xyz/objkt/164149 breaks when you click collectors
fix: encodeURI(swap.creator.name)

also changes other buttons from `href` to `to`


kylegrover:
I think we should also remove encodeURI from the displayed names if we can, it's super ugly lol. was it really a security problem?

crzypatchwork:
ppl were trying to impersonate others so it was kinda of needed

PR #1013, opened 2021-07-22
Change price of swapped edition

nftbiker:
This patch allow a collector to change the price of swapped editions of an objkt.
Currently, to change a price, you have to unswap, wait for tx to complete, then swap at new price.
This patch do the same thing but in a batch that cancel current swap and do a new swap with new price provided by collector through UI.

It can be tested on a temporary instance run at https://alterhen.netlify.app/
I already used multiple time without problem.
Attached is a screenshot of the reswap UI
reswap

### DIFF
--- a/src/components/owner-swaps/index.js
+++ b/src/components/owner-swaps/index.js
@@ -7,9 +7,9 @@ const sortByPrice = (a, b) => {
   return Number(a.xtz_per_objkt) - Number(b.xtz_per_objkt)
 }
 
-export const OwnerSwaps = ({ swaps, handleCollect, acc, cancel }) => {
-  swaps = swaps.filter(e => parseInt(e.contract_version) === 2 && parseInt(e.status) === 0 && e.is_valid )
-  console.log('v2',swaps)
+export const OwnerSwaps = ({ swaps, handleCollect, acc, cancel, reswapv2 }) => {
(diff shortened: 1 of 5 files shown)


powersurge01:
Looks very nice! It seems that it handles the UI and the the contract call via new "reswapV2" function

crzypatchwork:
maybe this could be added to another branch with some other functionalities. this feature is mostly wanted as a migration tool actually for reswapping from v1 -> v2

nftbiker:
As you wish, but it's a real collector need to be able to change price of what we have swapped. We, collectors, do this multiple times a day, either to decrease or increase price of what we sell, and having to unswap...wait...swap...wait is a pain. That's why this patch was build, to ease changing price. But it could indeed easily be adapted to build an automatic tool that just cancel v1 swap and reswap with same price/quantity to v2 contract. So for me it's 2 separate things, and the need to reswap v1->v2 don't change the fact that for collectors, being able to change price in one click is also a real thing.

kylegrover:
I agree with biker this is a great change already and would be good to get it live before thinking about reusing this work for additional features like v1 to v2 swap

peterschmidler:
I am using @nftbiker's tool under https://henplus.netlify.app regularly without any problems to change prices of pieces I want to sell on the secondary market exclusively. It's a great time saver and I wish you guys would merge it.

PR #1019, opened 2021-07-26
Feature - Add Discord Verification From Tezos Profiles

obstropolos:
This change adds a Discord badge to a creator's page if they have an existing Discord Verification created using the [Tezos Profiles](https://tzprofiles.com) service. The data is currently being pulled from the Tezos Profiles API.

Since Discord has no concept of a 'user page,' the badge on hover shows the user's handle, and when clicked, copies it to the clipboard. That handle can then be taken to the Discord app if they wish to message that user. 

A demo deployment of the service with the add-on can be found [here](https://4aa8a67d.hicetnunc.pages.dev/).

**Testing**
- Example of an existing profile with a [Discord verification](https://4aa8a67d.hicetnunc.pages.dev/tz/tz2Dvv6ib5wbtWkPghE2KdgHB9WXvoNxERQ8/)

Screen Shot 2021-07-26 at 12 23 42 PM

**Future Goals**
We will continually augment the Tezos Profiles service with additional self-serve verifications for creators in order to better verify themselves. As always, we would love any input on priority verifications to better serve the community.


kylegrover:
This is sweet! Something about the animation does feel a bit out of place to me though, between the fast moving sideways slide in and the slow click animation. Maybe we can just try to replicate the curate button tooltip? A 'Copied' indication and/or clipboard icon would help with UX as well.
Stoked for the feature for sure.

w4ll3:
Got it, instead of the animation I changed so that the tooltip becomes copied for 1 second and after that it returns to it's previous value (discord handle). If you have more feedback please do not hesitate.

kylegrover:
I think it's good to go. @crzypatchwork is the develop branch active or should this be moved to main?

violetforest:
i would suggest the animation that is on the hDAO curate button to keep things consistent

w4ll3:
are you referring to [this](/hicetnunc2000/hicetnunc/tree/main/src/components/item-info/styles.module.scss#L77)?

PR #1056, opened 2021-08-07
Update w.json

ThePartyy:
in god we trust copymint


ThePartyy (commented):
added another copymint

PR #1084, opened 2021-08-15
Added P5 SVG Template

gschian0:


gschian0:
I added a template to run P5js as an animated SVG to protect against the anti-aliasing demon and did a test mint I didn't upload that looked great! Can be done a little more heavy duty... but this cool little library... https://github.com/zenozeng/p5.js- is awesome to use for quick and easy!

PR #1100, opened 2021-08-22
Add TezTools Support 

ghost:
Adding teztools data and media support, currently our main source is data however we also provide media to correct broken IPFS links that token creators added to metadata.


IntegroLabs:
Also fonts and other support are added for future NFT features on our roadmap

AlexandreRangel:
Any news on this :) ?

PR #1108, opened 2021-08-26
Add the ZIM JavaScript Canvas Framework Interactive NFT Template

danzen:
We have been successful minting interactive NFTs with ZIM using a similar template to the one provided.  Thank you hic et nunc for this opportunity to spread creative coding!


danzen:
I think we may have done this wrong... we forked from the main repository and then committed to develop.  Let us try again.  It was to add a directory to the template directory.  Would you still like us to fork from develop?

PR #1182, opened 2021-09-19
copy wallet button on artist and objkt pages

kylegrover:
quick access to the full wallet, to paste into other apps ( and to make it faster for me to ban people :P )
image
image


kylegrover:
image

PR #1297, opened 2021-10-15
update markdown mimetype

fraguada:
from https://stackoverflow.com/questions/10701983/what-is-the-mime-type-for-markdown


crzypatchwork:
what happens is that when a text/markdown gets uploaded it's returning an unknown mimetype, probably related to FileReader? https://github.com/hicetnunc2000/hicetnunc/blob/312b52488f03a45763db567329e6cdb6c0fcea42/src/components/upload/index.js#L17

PR #1330, opened 2021-11-07
Bugfix/stop console errors

TezzardzGoneWild:
The console has a few errors on profile page load. This removes the errors coming from the below:
- one of the SVGs had multiple properties using dashes instead of camelCase (ex. `stroke-width` instead of `strokeWidth`), and React doesn't read these properly. Same with using `class` instead of using `className`
- two functions mapping arrays to HTML elements had the key on one of the children instead of the parent -- React wants keys on the topmost element when something is mapped

An additional bug was fixed:
-  this.state.alias is mentioned multiple times, but it does exist. This was improperly used causing every tab to have the title `hic et nunc - hic et nunc` instead of the profile's name (e.g. `tezzardz_gone_wild - hic et nunc`). This reference has been replaced with `this.state.subjkt` to fix the bug and other references removed

Several sections of code received some clean up, but no functionality change:
- `onReady` included a nested if statement that could be expressed using a single if statement
- the nested if statement used for filtering OBJKT sale type within `creationsForSale` can be expressed with two one-line if statements
- metadata fields are  set to `null` when the state is initialized to give a better indication of what will happen within the component
- instances where `getUserMetadata` is called wrote to an intermediate variable and had multiple unnecessary if statements where writing `null` or `undefined` would have the same outcome
- `this.state.copied` has been renamed to `this.state.discordLinkCopied` for clarity

### DIFF
--- a/src/pages/display/index.js
+++ b/src/pages/display/index.js
@@ -607,15 +607,9 @@ export default class Display extends Component {
               <Identicon address={this.state.wallet} logo={this.state.identicon} />
 
               <div className={styles.info}>
-                {this.state.alias && !this.state.subjkt ? (
-                  <p>
-                    <strong>{this.state.alias}</strong>
-                  </p>
-                ) : (
-                  <p>
-                    <strong>{this.state.subjkt}</strong>
-                  </p>
-                )}
+                <p>
+                  <strong>{this.state.subjkt}</strong>
+                </p>
                 {this.state.description && <p>{this.state.description}</p>}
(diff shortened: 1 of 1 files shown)


TezzardzGoneWild:
In favor of smaller PRs, this has been closed and replaced by these:
- https://github.com/hicetnunc2000/hicetnunc/pull/1331
- https://github.com/hicetnunc2000/hicetnunc/pull/1332
- https://github.com/hicetnunc2000/hicetnunc/pull/1333
- https://github.com/hicetnunc2000/hicetnunc/pull/1334

